PR #7, opened 2015-02-03
Update README 'Quick Install' snippet to load the branch SDK via https

theunixbeard:
...rather than http. This prevents mixed content errors on https sites using the 'Quick Install' snippet in modern browsers such as Chrome v40

branch_api_mixed_content_error


dmitrig01:
Yeah - thanks!

PR #8, opened 2015-02-06
Version 2

scotthasbrouck:


src/0_utils.js
```diff
@@ -51,7 +51,7 @@ utils.readStore = function() {
 		return JSON.parse(sessionStorage.getItem('branch_session')) || {};
 	}
 	catch (e) {
-		return {};
+		return {}; //TODO: Look into why we dont return the error
```
dmitrig01: Hrm, the likelihood that there's an error here is incredibly small - someone would have to screw with the sessionstorage. which is possible, but in that cas, we can just cancel it and do a new open. Make sense?

scotthasbrouck: Yep, makes sense

src/2_branch.js
```diff
@@ -93,42 +116,52 @@ Branch.prototype['logout'] = function(callback) {
...
 		}, {})
-	}, callback);
+	}, callback({}));
 };
 
 /**
+ * @param {string} identity 
+ * @param {?function} callback
  */
 Branch.prototype['identify'] = function(identity, callback) {
```
dmitrig01: It'd be great to keep the calling convention here as I had it before - since identity now refers to an object that only will contain one property, "identity", we might as well just pass that in as a string. Does that make sense?

dmitrig01:
Anywhere you pass in `app_id`, `session_id`, or `identity_id` to the `api` function: it should actually fill those fields in automatically - see https://github.com/BranchMetrics/Web-SDK/blob/v2/src/2_branch.js#L22-L26

src/2_branch.js
scotthasbrouck: Yes, that makes perfect sense. I thought that seemed strange, I was just afraid of breaking existing integrations of the WebSDK out in the wild, since the current version takes an object. I suppose I could check if it is passed an object, and pull the string out of it to make it backwards compatible.

src/0_utils.js
```diff
@@ -8,7 +8,7 @@ goog.provide('utils');
 var DEBUG = true;
 
 /** @typedef {string} */
-message;
+var message;
```
dmitrig01: Ah, I think this should actually stay as-is - the idea here is to define a type to the closure compiler, not to actually initialize a variable

src/0_utils.js
```diff
@@ -74,20 +74,6 @@ utils.hashValue = function(key) {
...
-	return navigator.userAgent.match(/android|i(os|p(hone|od|ad))/i);
-};
-
-// TODO: this probably should not be a utility
-utils.closeBanner = function() {
-	var d = document.getElementById('branch-banner');
-	if (d) {
-		d.parentNode.removeChild(d);
-		document.body.style.marginTop = '0px';
+		return ''; //TODO: Look into why we dont return the error
```
dmitrig01: I don't know if this would really even throw an error :P this is just code adopted from before

src/2_branch.js
```diff
@@ -322,6 +357,40 @@ this.appBanner = function(obj) {
 				phone.className = 'error';
 			}
 		};
+
+	// Close banner
+	var closeBrancheBanner = function() {
```
dmitrig01: typo - branche :-)

src/2_branch.js
dmitrig01: Ah, there a ton of backwards-incompatible changes - the biggest one being the callback parameter change. I'd rather just break everything in one go and document an upgrade path.

src/2_branch.js
```diff
@@ -74,16 +73,40 @@ Branch.prototype['init'] = function(app_id, callback) {
...
+			utils.store(data);
+			callback(data);
+		}
+	});
+};
+
+/**
+ * @param {function|null} callback
+ */
+Branch.prototype['close'] = function(callback) {
```
dmitrig01: Hrm, thinking about this - does it even make sense to have this function? not sure why someone would possibly want to use it

dmitrig01: perhaps we can think about it later - just don't waste too much time ont his one :P

src/2_branch.js
scotthasbrouck: Alright, sounds good!

src/2_branch.js
```diff
@@ -139,182 +172,184 @@ Branch.prototype['createLink'] = function(obj, callback) {
...
+/**
+ * @param {?Object} metadata
+ * @param {?function} callback
+ */
+Branch.prototype['SMSLink'] = function(obj, callback) {
+	callback = callback || function() {};
+	if (!this.initialized) { return utils.console(utils.messages.nonInit); }
+	obj["channel"] = 'sms';
+	var self = this;
+	this.createLink(obj, function(url) {
```
dmitrig01: So previously, and this behavior was quite confusing, this function checked if you had clicked a link to the page, and if so, it actually sends the same link, not a new one. I would like to actually preserve that functionality, as it's pretty widely used – however, it's very confusing; perhaps we could split out into three functions like `createAndSendLink`, `sendExistingLink`, and `sendExistingOrNewLink`? Except with better names :P

src/2_branch.js
```diff
@@ -139,182 +172,184 @@ Branch.prototype['createLink'] = function(obj, callback) {
...
-	self.api.makeRequest(config.resources.referrals.showReferrals, {
-		app_id: config.appId,
-		identity_id: utils.identity()
-	}, function(data) {
-		if (typeof callback == 'function') { callback(data); }
+
+/**
+ * @param {?function} callback
+ */
+Branch.prototype["showReferrals"] = function(callback) {
```
dmitrig01: we can probalby rename a bunch of these - like this would probably just be better called referrals, the next would probably be better called credits, etc. The goal is to make this feel as idiomatic as possible for people using the library – it needs to feel like any other JS library

src/2_branch_tests.js
```diff
@@ -0,0 +1,379 @@
...
+		},
+		asyncPollingInterval,
+		maxWaitTime
+	);
+	
+	branch.SMSLink(params, function(data) {
+		recievedFired = true;
+		recievedData = data;
+	});
+};
```
dmitrig01: It looks like you're repeating the same structure in all of these tests - perhaps we can abstract this out into a helper function, so you can call it something like this:
```
var testSMSLink = function() {
    runTest(function(cb) {
        branch.SMSLink(params, function(data) { cb(data); }
    }, function(data) {
        assertObjectEquals("should send SMS with phone number", data, {});
    });
};
```
(i.e. first param is "do the thing" function, and second is assertions function)

src/2_branch.js
scotthasbrouck: Ahhhh, Ok. Yes. I'll break that apart into 3 functions - that'll be way less confusing, and it certainly makes sense to just reuse the same link that was already there. Just to be clear, all three functions would be external? Not sure if there is a common use case where a new link would need to be created if they already clicked a link

src/2_branch.js
scotthasbrouck: Yeah, I was already there. I don't like all the verbs in these function names. But if we're going to break past integrations, then let's just do it.

src/2_branch.js
dmitrig01: Well, the idea is, if the user clicked a link, use that, but users didn't always click a link, so if they didn't, just create a new one. But I could see also falling back to some link you pass in instead of creating a new one every time. Does that make sense?

src/2_branch_tests.js
scotthasbrouck: Good call. Yes, the setup is nearly identical in all of them. That will cut out a lot

dmitrig01:
Hey let's split out tests and test-related files into their own subdirectory, `test/`. For the autogenerated file (calcdeps.py?) - can you make some way to regenerate it - maybe in the makefile? Happy to help with makefile syntax if needed

dmitrig01:
Let's add something in the makefile also to download closure-library for tests. And then let's put all the autogenerated files in gitignore :-)

src/2_branch.js
dmitrig01: awesome! let's break it all, make sure it's perfect, and then document an upgrade path (we can hold people's hand through it too if they really need it)

src/branch_tests.html
```diff
@@ -0,0 +1,22 @@
...
+  <script src="../../closure-library/closure/goog/base.js"></script>
+  <script src="../../closure-library/closure/goog/testing/testcase.js"></script>
+  <script src="../../closure-library/closure/goog/testing/testrunner.js"></script>
+  <script src="../../closure-library/closure/goog/testing/propertyreplacer.js"></script>
+  <script src="../../closure-library/closure/goog/testing/continuationtestcase.js"></script>
+  <script src="branch-deps.js"></script>
+  <script src="2_branch_tests.js"></script>
+  <script>
+  	goog.require('Branch');
+    goog.require('utils');
```
dmitrig01: Hrm, any reason to require these two specifically? Or actually - how does the 2_branch_tests.js get its dependencies loaded? just closure library magic? :P

scotthasbrouck:
Will do!

src/branch_tests.html
scotthasbrouck: You know what, I don't need utils - not sure why I put it there :-P
and ya, 2_branch_tests is magic!

src/2_branch_tests.js
scotthasbrouck: Taken care of in 00cc6b0856bf9cadf93a5d71b1d8db3dba0af207

scotthasbrouck:
I broke out the banner into 1_elements.js and refactored it - it's so huge I think it needs to be in it's own file. It's also been restyled to match iOS 8 and has transition animations. The thought with naming the file elements is that any future UI elements we add to the Web SDK could go in that file (perhaps we could add UI's for referrals, etc). But after building it out, it may make more sense to just rename it banner.js, and make a new file for any future UI features.

src/2_branch.js
scotthasbrouck: Agreed, I can't see any possible reason for needing to do this. I haven't done anything with it yet though. Perhaps get some input from a few others? There just may be edge cases I'm not aware of.

dmitrig01:
doing some review

### readme (and API I guess):

- [ ] several typos: "Requriements", "app_di", "queed", "accountrmation", "imrpovements", "wether" (this was a more goaty product, but alas...) etc. let's get someone to read over this and spell/grammar/make-sense check when it's done - maybe @derrickstaten?
- [ ] very minor, but the Xes look ilke we don't support it; perhaps a check mark? :-)
- [ ] some extreme/funky indentation, no?
- [ ] most of these functions return nothing - what you classify as "returns" is actually the data passed to the callback, right?
- [ ] I know the server-side call is named `profile` but I think it would make more sense as `identify` (like Mixpanel) or `setIdentity` (like our SDKs)
- [ ] not sure how much control we have over jsdoc formatting, but "'init'(app_id, callback)" looks weird. There are a bunch of ways of denoting this; the haskell/loosely-category-theoretical way would be something like `init :: app_id, ((err, data) -> null) -> null` but not sure if this is the most intuitive way for JS developers. Let's find some good reference docs out there and copy them
- [ ] `Note: Branch.init is called every time the constructor is loaded.` this doesn't make sense - what do you mean by "constructor is loaded"? Are you trying to say `Note: you must call Branch.init before using any other Branch SDK method`?
- [ ] for `init` let's be more clear that "this is where you can read the data about the link the user used to get to this page". i'm not sure we need to return all that data to the user - let's only return them stuff they need: data and referring_identity
- [ ] on the profile call - let's include a little more info on what exactly the identity string is for
- [ ] maybe we should have some kind of more overviewey section that says something along the lines of "to properly work with user tracking, you will have to call our SDK on login and logout events; do that with `identify` and `logout`"
- [ ] still not sure of a use-case for `close` - i'd say let's remove it (or leave it in but commented out)
- [ ] for `link` - let's "return" just the link, not that object. call the parameter `data` not `metadata`. let's include some docs on what that parameter is - we can copy from the iOS or Android guides, it's pretty well documented
- [ ] i might keep `event` as `track`, as it's in line with what mixpanel does - what do you think?
- [ ] `SMSLink` - still not sure on the name - how about `sendSMS`? And let's be super clear in the description: *if the user got to this page from a branch link, send that same link; otherwise, create a new link with the data provided here*. would be great to make the link parameter optional too, and if you omit it, it just sends a link with everything default and channel SMS...
- [ ] would be great to be able to pass `SMSLink` a string (of the form `bnc.lt/l/XYZ` or some kind of white-labelled link [for now, we can just assume it's a branch link? can also validate it's the form `http(s?)://(domain name)/l/[A-Za-z0-9_\-]+`) so it doesn't create a new link every time. maybe this is scope creep for now?
- [ ] `SMSLinkNew` - also unsure about the name. now that I'm seeing it laid out here, I think it makes more sense as an optional boolean flag to sendSMS: `use_referring_link_if_exists` or something ugly like that, defaulting to true? (but optional) (same unease for `SMSLinkExisting`; not sure if this function makes sense as a standalone)
- [ ] `credits` and `referrals` - i know this is a bit redundant, but can we copy some text on what these actually show from the iOS or Android docs?
- [ ] `redeem` - let's pass it two params, `bucket` and `amount`. looks like it's not documented as getting (err, data)?
- [ ] `banner` - love it. would love to be more thorough on docs here:
  - [ ] screenshot!
  - [ ] let's pull this data from the branch dashboard if you've set it - i'll work on the server-side piece to do that
  - [ ] let's document better what these parameters are. it looks like the data sub-object _within_ this object is the link parmaeters (and that itself can have a data object)? what about splitting out this into a second parameter, with a similar format to SMS - you can pass a link data, link string, or nothing? and it can make the channel 'app banner' by default...
  - [ ] does the callback get `data` or `err, data`? is the callback useful to the developer?
  - [ ] are we going to have options for mobile-only, mobile-and-desktop, desktop-only?

### Overall:

- [ ] we used to have branch.utils.readSession() which people would use to get the referring data. we sholud probably provide some method to replace that. it would be awesome to make it an async method, so the function would be like: load data if it's not there (i.e. waiting for an init call - really, the same as an init call, but without the app id, because init has to already have been called) or return it if it's there
- [ ] we need a request queueing system - i can fill in more details on this - but the gist is: certain calls (init, identify, logout) can change the identity_id and/or session id; if you do something like `branch.track('a'); branch.logout(); branch.track('b');`, we want the second event to get tracked with the post-logout session id. does that make sense? there are a couple of ways of doing this. the easiest way is probably just having a queue of requests so when you call `this.api` in the branch object, instead of executing, it adds to the queue (or executes immediately if the queue is empty); when execution finishes, it calls the next item on the queue. that's why we add in identity_id, session_id, and app_id _separately_ - because it lets us do this very easily
- [ ] some code formatting, styling, indentation, etc. inconsistencies. Looks like I have jshint in gulp, we should make sure to run that. I'll add JSCS as well.

### Code specifics

- [ ] `1_elements.js` maybe better named `1_banner`?
- [ ] let's take `Branch.prototype.api` off the branch prototype since it's an internal function
- [ ] let's make sure to add back IE9 and IE10 specific fixes
- [ ] doesn't your `bannerShown` actually mean `bannerHidden`?
- [ ] can you clarify your comment `message; //Does not work with tests`? let's work this out in chat
- [ ] `referrals` and `credits` - would be great to cache these client-side (we can do it per-page load right now, not in the session cache)
- [ ] i saw you updated the on-page script; let's update the source of the script - `src/onpage.js` - and have a build step which minifies it puts this in the docs

### Tests:

- [ ] for testing - the tests are good as-is, but I would like to see another type of test as well: it would be great to stub out server calls so we can: (a) make sure requests are sent to the server in the proper format, and (b) simulate different responses. Specifically, to do this, we would need:
  - [ ] In `1_api`, separate the request-building part (https://github.com/BranchMetrics/Web-SDK/blob/implements_sms/src/1_api.js#L85-L99) from the XHR part (rest of that function).
  - [ ] Add a mocked/stubbed XHR-runner function
  - [ ] start to write some basic tests that the SDK is actually sending requests in the right format; simulate some errors as well (invalid app ID, for example) and see how they are passed back
  - [ ] more testing along that line: would be great to use this to test the request queueing type thing I mentioned above, as we can do exactly the calls i had above, and ensure the second event is tracked with the second session id (fake-returned by our mocked logout api call).
- [ ] let's add some docs or automated tooling around running tests. those relative paths in tests/branch_tests.js look janky :-)

### Release:

- [ ] let's write up a release process, (and even think about turning into a script?!)
- [ ] let's release it as 1.0.0!

scotthasbrouck:
- [ ] Specify call back format as cb(err), or cb(err, res)
- [ ] Don't return empty objects
- [ ] sendSMS & banner, and all other functions, phone, linkData, options object
- [ ] Make sure callback is last param on all
- [ ] redeem, two params int and string

scotthasbrouck:
Release Procedure
- [ ] JSHint one last time
- [ ] Check example.html with README, make necessary updates
- [ ] Upload gzip to Amazon S3
- [ ] Update include snippet in README and example.html with Amazon S3 URL
- [ ] Update Changelog to today's date, commit as tag for version 1.0.0
- [ ] Send email, and post blog! And all the fun social markety things
------------
- [ ] Bower, npm

src/2_branch.js
```diff
@@ -6,322 +6,740 @@ goog.provide('Branch');
...
+ * );
+ * ```
+ *
+ * ## Credit redemption
+ *
+ */
+Branch.prototype['credits'] = function(callback) {
+	callback = callback || function() {};
+	if (!this.initialized) { 
+		this.nextQueue();
```
dmitrig01: hrm, this doesn't exist any more right?

src/2_branch.js
```diff
@@ -6,322 +6,740 @@ goog.provide('Branch');
...
+ *
+ * The Branch Web SDK has a built in sharing banner, that automatically displays a device specific banner for desktop, iOS, and Android. If the banner is shown on a desktop, a form for sending yourself the download link via SMS is shown.
+ * Otherwise, a button is shown that either says an "open" app phrase, or a "download" app phrase, based on whether or not the user has the app installed. Both of these phrases can be specified in the parameters when calling the banner function.
+ * **Styling**: The banner automatically styles itself based on if it is being shown on the desktop, iOS, or Android.
+ *
+ */
+Branch.prototype['redeem'] = function(amount, bucket, callback) {
+	callback = callback || function() {};
+	if (!this.initialized) { 
+		this.nextQueue();
```
dmitrig01: nor this

src/2_branch.js
```diff
@@ -6,322 +6,740 @@ goog.provide('Branch');
...
+ * `Branch.init`, but does not require the `app_id`. This is meant to be called
+ * after `Branch.init` has been called if you need the session information at a
+ * later point.
+ * If the Branch session has already been initialized, the callback will return
+ * immediately, otherwise, it will return once Branch has been initialized.
+ * ___
+ */
+Branch.prototype['data'] = function(callback) {
+	callback = callback || function() { };
+	callback(null, utils.whiteListSessionData(utils.readStore()));
```
dmitrig01: let's put this inside the queue, so that if you do branch.init() branch.data() it calls data *after* init finishes

scotthasbrouck:
Cool, data() is in the queue now, and those two relic nextQueue calls are gone, not sure how they snuck thru.

PR #12, opened 2015-02-26
Fingerprint from get param with fallback to hash

scotthasbrouck:
@dmitrig01 grabs r from url. Preference to get param ?r=, then falls back to #r:, or returns undefined.

### DIFF
--- a/src/branch.js
+++ b/src/branch.js
@@ -42,15 +42,24 @@ var Branch = function Branch(app_id, debug, callback) {
 		return fin;
 	};
 	this.utils.hashValue = function(key) {
-		var v;
 		try {
-			v = location.hash.match(new RegExp(key + ':([^&]*)'))[1];
+			return location.hash.match(new RegExp(key + ':([^&]*)'))[1];
 		}
 		catch (e) {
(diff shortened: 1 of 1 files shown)


dmitrig01:
oh sweet - I was thinking that we would just put this in the sdk v2, but i guess there's no reason not to do that. I'd like to name the new parameter something more descriptive. Maybe `branch_id`? But perhaps that's prone to collision. Any other ideas?

scotthasbrouck:
Yeah, it was so simple, I figured I would just do it here in case we needed/wanted it. Agreed, a more descriptive parame name makes more sense. I have a break in a few hours for the rest of the afternoon, I'll give it some thought then

scotthasbrouck:
Oh question, any reason to keep the hash as just 'r'? Perhaps we can make that more descriptive, also.

dmitrig01:
i think we need to keep that for backward-compatibility

dmitrig01:
gonna close this, i think that's ok. we could do it against a legacy branch, but hopefully we can phase out 0.x soon enough

PR #21, opened 2015-03-22
Append class name to preserve existing classes.

IsaiahJTurner:


dmitrig01:
Ah - good find - thanks.

PR #26, opened 2015-04-07
Simple fix for updating web sdk src in readme during release

scotthasbrouck:
This seems to work for me, and includes the correct src tag in README when I run make or ./release.sh. @dmitrig01 you're more familiar with Make than me, does this seem like the correct fix?


dmitrig01:
hrm, so i think the issue is what happens on subsequent makes - namely that the readme gets rebuilt, but since there's no VERSION passed in, onpage defaults to dist/build.js

I think we need to fix two things here:
- right now, onpage ends up in two places: the README and the example.html. we want README to point to the latest release, whereas we want example.html to point to dist/build.js in non-release builds; thus it probably doesn't make sense to build onpage into its own file any more - right?
- in order, on every make (even a non-release make) to get the right version to put in the readme, we'd need to read it from some external source. probably easiest is package.json - this will pull it out for you: `grep "version" package.json | perl -pe 's/\s+"version": "(.*)",/$1/'`

scotthasbrouck:
ah, both of those points make sense, because I was assuming only running ./release.sh vx.x.x and make version=vx.x.x

Correct. I can't think of a scenario where we want dist/build.js in the Readme, and we certainly want example.html to point to the local build when not releasing. So yes, I think it makes more sense to not build onpage into it's own file.

Reading from either config.js or package.json would work. I'll just build the init code separately for example.html and README.md and print the init with proper src directly into them.

dmitrig01:
wow man you now know more about makefiles than i do

PR #30, opened 2015-04-10
Banner on init and closeBanner

scotthasbrouck:


scotthasbrouck:
@dmitrig01 Thoughts on this for https://github.com/BranchMetrics/Web-SDK/issues/29

dmitrig01:
hrm, to be honest, I'm not sure what the benefit of putting it in init is over having two function calls

the one thing that would be gerat to fix, though, is the queueing, so that you could just do:
```javascript
branch.init('key');
branch.banner(...);
```

instead of
```javascript
branch.init('key', function() { branch.banner(...); });
```

dmitrig01:
ping on this :-)

scotthasbrouck:
Everything in this PR:
- Fixes bug with app banner that caused the SMS waiting indicator to be misplaced
- Correctly implements queue so that this works, for example:
```js
branch.init();
branch.banner();
```
- Updates Cordova guide and adds upgrade guide. Both guides get the up-to-date init code via the makefile, just like the readme.
- Switches Web SDK to using API key rather than App ID, with automatic detection and fallback of App id
- adds depreciation notes to readme and cordova guides re: app id
- Adds section to readme explaining the queue and how it can be used to call the asynchronous sdk methods back to back, and that they will execute in order and wait for their turn, basically.
- Adds `closeBanner()` method

scotthasbrouck:
@dmitrig01 all set for review

src/1_resources.js
```diff
@@ -27,14 +27,17 @@ var _validator;
...
  * @param {boolean} required
  * @param {validationTypes|RegExp} type
+ * @param {string=} fallback
+ * @param {validationTypes|RegExp=} fallbackType
  * @throws {Error}
- * @return {_validator}
  */
-function validator(required, type) {
-	return function(endpoint, param, data) {
+function validator(required, type, fallback, fallbackType) {
```
dmitrig01: hrm, i know this is a bit of a cop-out, but perhaps it would be cleaner to do this at the application layer (make app_id and branch_key optional, and in api.js, verify that at least one of them is present)?

dmitrig01: er not "applciation layer", but one layer above

dmitrig01: that way we don't need to modify all these resources, plus the cordova ones need to change too :P

src/2_banner.js
```diff
@@ -144,11 +144,15 @@ banner = function(branch, options, linkData, storage) {
 				sendSMS(doc, branch, options, linkData);
 			});
 		}
+
 		var closeButton = doc.getElementById('branch-banner-close');
+		banner.close = function() {
```
dmitrig01: hrmm, seems a little weird to set a new method on banner like this. not sure how to rectify though. somewhere we gotta store a pointer to the element. maybe just return this function, and the branch object can store this function and retrieve it later? or maybe that is not as clean?

CORDOVA_GUIDE.md
```diff
@@ -0,0 +1,430 @@
...
+Called when an app first initializes a session, ideally in the app delegate. If you created a custom link with your own custom dictionary data, you probably want to know when the user session init finishes, so you can check that data. Think of this callback as your "deep link router." If your app opens with some data, you want to route the user depending on the data you passed in. Otherwise, send them to a generic install flow.
+
+This deep link routing callback is called 100% of the time on init, with your link params or an empty dictionary if none present.
+
+**[Formerly `getInstance() and initSession()`](CORDOVA_UPGRADE_GUIDE.md)**
+
+
+```js
+// adds an instance of branch to the window object
+(function(b,r,a,n,c,h,_,s,d,k){if(!b[n]||!b[n]._q){for(;s<_.length;)c(h,_[s++]);d=r.createElement(a);d.async=1;d.src="https://cdn.branch.io/branch-v1.3.4.min.js";k=r.getElementsByTagName(a)[0];k.parentNode.insertBefore(d,k);b[n]=h}})(window,document,"script","branch",function(b,r){b[r]=function(){b._q.push([r,arguments])}},{_q:[],_v:1},"init data setIdentity logout track link sendSMS referrals credits redeem banner closeBanner".split(" "),0);
```
dmitrig01: fyi - for cordova, due to some stupidity, it's always going to be there, and always global - no need to put the snippet in

CORDOVA_GUIDE.md
```diff
@@ -0,0 +1,430 @@
...
+
+// Arguments
+// arg1: Your API key can be retrieved on the [Settings](https://dashboard.branch.io/#/settings) page of the dashboard
+// arg2: the callback to notify you that the instance has instantiated
+branch.init("API-KEY", function(err, data) {
```
dmitrig01: let's always, always, always call it "Branch key", never API key or App key :-)

CORDOVA_UPGRADE_GUIDE.md
```diff
@@ -0,0 +1,235 @@
+# Upgrade Guide To The New Branch PhoneGap/Cordova SDK
+
+In the effort of making the quickly growing functionality of our Web SDK, available to your PhoneGap or Cordova apps, we've merged the two together. This will require some very simple changes to the implementation, but you'll gain a number of very useful functions and the Branch promise of continuous updates and improvements!
```
dmitrig01: what about a table with two columns - "before" and "after"?

README.md
```diff
@@ -23,26 +23,24 @@ This SDK requires native browser Javascript and has been tested in all modern br
 | ------ | ------- | ------ | ---------- |
 |    &#10004;   |    &#10004;    |   &#10004;    |  9, 10, 11 |
 
-### API Key
+### API Key (formerly App ID)
```
dmitrig01: BRANCH KEY ;-)

CORDOVA_UPGRADE_GUIDE.md
scotthasbrouck: Yes. MUCH cleaner.

README.md
scotthasbrouck: Haha, I'll do a txt search thru the whole code base for anything api-key, api key, etc

src/3_branch.js
```diff
@@ -176,101 +178,112 @@ if (config.CORDOVA_BUILD) {
  * **Note:** `Branch.init` must be called prior to calling any other Branch functions.
  * ___
  */
-Branch.prototype['init'] = function(app_id, options, callback) {
-	if (options && typeof options == 'function') {
-		callback = options;
-		options = { isReferrable: null };
+Branch.prototype['init'] = function(key_or_id, options, callback) {
```
dmitrig01: let's just call the parameter branch_key -- and include in the comment that you can optionally use your app id, but it's deprecated

README.md
dmitrig01: :+1:

dmitrig01:
hrm, two overarching things:

1. I think there may still be an issue in init: if you call `branch.init(); branch.track();`, branch.track will fail because it hasn't been initialized; see https://app.asana.com/0/17126854785893/30218910917492
2. I think we can clean up the queueing code, initialization code, AND error handling code by refactoring the wrapError functions into just one function, a `wrap` function, that works like this:

```
branch.track = wrap(1, function(done, event, metadata) {
	this._api(resources.event, {
		"event": event,
		"metadata": utils.merge({
			"url": document.URL,
			"user_agent": navigator.userAgent,
			"language": navigator.language
		}, metadata || {})
	}, done);
});
```

`wrap` has 2 parameters:
- how many parameters the callback takes (0 = no callback; this is useful for .banner; 1 = just an error param; 2 = error, data... actually we sholud probably use constants for this: NO_CALLBACK, CALLBACK_ERR, CALLBACK_ERR_DATA)
- a function which actually does the stuff; this gets executed:
  - after init has completed (if init hasn't started, either pass the error into the callback if there is one, or throw it)
  - in its proper place in the queue

the function receives the callback parameter first, if it's not NO_CALLBACK; if has a callback, but the callback is passed in (if the third parameter is not NO_CALLBACK), the final parameter to the function is considered to be the callback, assuming it's a function - if it's not, an empty function is passed in, which console.logs and throws errors

does this make sense?

dmitrig01:
also on #1, it's possible to call branch.banner now without calling branch.init since there's no validation (we should probably have a test for that too :-))

src/3_branch.js
```diff
@@ -1123,27 +1181,52 @@ if (config.WEB_BUILD) {
...
+				bannerOptions.showiOS = bannerOptions.showAndroid = options['showMobile'];
+			}
 
-		banner(this, bannerOptions, linkData, this._storage);
+			banner(self, bannerOptions, linkData, self._storage);
+			next();
+		});
+	};
+
+	Branch.prototype['closeBanner'] = function() {
```
dmitrig01: hrm, i think there may be an argument for having this always defined, and just fail silently when the banner is open... but i could be wrong

src/2_banner.js
scotthasbrouck: I like the idea of the banner just returning the close method. If we want the banner to have other methods in the future, we can just have the banner return an object of functions, and store that object on branch. I'll just have it return the close function for now.

dmitrig01: yea sounds great :-)

CORDOVA_UPGRADE_GUIDE.md
scotthasbrouck: Wow, so this is impressively more difficult than you'd think: putting a code snippet inside a table w/ markdown. Going to revisit after everything below.

src/3_branch.js
scotthasbrouck: Agh, you're correct on number 1. I had originally removed the nonInit error throws,  and at that point, the queue did work with every function. When I put them back in, I didn't think to put them back inside the queue.
With that said, yes, all of this makes sense and it sounds like a great idea. I'll get to work on it.

CORDOVA_UPGRADE_GUIDE.md
dmitrig01: hah, just googled it, apparently it's not possible. Maybe it's worth just doing it like:

```
### Initializing the Branch SDK
#### Old:
XXX
#### New:
XXX
----
### Closing the session
#### Old:
XXX
#### New:
XXX
----
```

dmitrig01: i guess that's not that different from what you have now, but it just makes it clearer... if you think it'd be good let's do it; if not, let's not waste time

scotthasbrouck:
Oh sweet! for sure.

src/3_branch.js
dmitrig01: ;-)

If you put it in wrap, it'll be super clean!

scotthasbrouck: And we like clean.

dmitrig01: mr _clean_logo

src/3_branch.js
```diff
@@ -285,15 +284,9 @@ Branch.prototype['init'] = function(app_id, options, callback) {
...
-	if (!callback) { return; }
-
-	var self = this;
-	this._queue(function(next) {
-		callback(null, utils.whiteListSessionData(utils.readStore(self._storage)));
-		next();
-	});
-};
+Branch.prototype['data'] = wrap(2, function(done) {
+	if (typeof done == "function") { done(null, utils.whiteListSessionData(utils.readStore(this._storage))); }
```
dmitrig01: we verify that it's always a function, right?

scotthasbrouck: Yes, I can't imagine how it could not be. Now I'm just trying to remember why I did this... Seems that the tests all still pass if I remove that validation. Your call.

scotthasbrouck:
Ok, I've got all passing tests. However, I had to comment out assert.throws() in basic tests. It turned into a time suck, and I'll need to revisit it. Let me know if you want me to get to the bottom of that before releasing this.

src/3_branch.js
dmitrig01: yeah let's remove it everywhere. part of the reason to have the wrap is to allow it

src/3_branch.js
```diff
@@ -285,15 +286,9 @@ Branch.prototype['init'] = function(app_id, options, callback) {
...
-Branch.prototype['data'] = function(callback) {
-	if (!callback) { return; }
-
-	var self = this;
-	this._queue(function(next) {
-		callback(null, utils.whiteListSessionData(utils.readStore(self._storage)));
-		next();
-	});
-};
+Branch.prototype['data'] = wrap(2, function(done) {
```
dmitrig01: let's also use the constants here -- if you put callback_params outside of wrap they can be accessed here, it's much clearer to read with a constant

dmitrig01: osrry - didn't realize this wasn't clear before :P

dmitrig01:
thanks! yeah, no need to keep digging, i'll take a look if i get a chancce

can you change API-KEY to branch_key?

dmitrig01:
or BRANCH_KEY?

src/0_utils.js
```diff
@@ -76,6 +76,17 @@ utils.whiteListSessionData = function(data) {
 	};
 };
 
+utils.cleanLinkData = function(linkData, config) {
+	if (config.WEB_BUILD) {
+		linkData['source'] = 'web-sdk';
+		if (linkData['data']['$desktop_url'] !== undefined) {
+			linkData['data']['$desktop_url'] = linkData['data']['$desktop_url'].replace(/#r:[a-z0-9-_]+$/i, '');
```
dmitrig01: sorry -- i know this is how it was before -- but gotta clean the branch_match_id out of here too

src/1_api.js
```diff
@@ -71,6 +71,22 @@ Server.prototype.getUrl = function(resource, data) {
 			}
 		}
 	}
+	// check for branch_key then app_id here
+	var branch_id = /^[0-9]{15,20}$/;
+	var branch_key = /((?:[a-z][a-z]*[0-9]+[a-z0-9]*))/;
```
dmitrig01: we can use a stricter regex here: `/key_(live|test)_[A-Za-z0-9]{32}/` is the actual pattern

src/1_resources.js
```diff
@@ -28,7 +28,6 @@ var _validator;
  * @param {boolean} required
  * @param {validationTypes|RegExp} type
  * @throws {Error}
- * @return {_validator}
```
dmitrig01: this should stay right?

src/1_resources.js
```diff
@@ -19,7 +19,7 @@ var validationTypes = { obj: 0, str: 1, num: 2, arr: 3, bool: 4 };  // now inclu
 
 /* jshint ignore:start */
 
-/** @typedef {function(string, string, *)} */
+/** @typedef {function(string, string, *, (Object|null)=)} */
```
dmitrig01: this change isn't needed ay more i think?

src/3_branch.js
```diff
@@ -16,47 +16,54 @@ if (config.CORDOVA_BUILD) { var exec = require("cordova/exec"); } // jshint igno
...
-	};
-}
+	/***
+	 * @param {Error} err
+	 * @param {function(?Error,?=)=} callback
+	 */
+	function wrapError(err, callback) {
+		if (callback) { return callback(err); }
+		throw err;
+	}
```
dmitrig01: this doesn't actually get used, right?

src/3_branch.js
```diff
@@ -16,47 +16,54 @@ if (config.CORDOVA_BUILD) { var exec = require("cordova/exec"); } // jshint igno
...
+				if (err) { throw(err); }
+			};
+			args = Array.prototype.slice.call(arguments);
+		}
+		else {
+			args = Array.prototype.slice.call(arguments, 0, arguments.length - 1) || [];
+			callback = lastArg;
+		}
+		self._queue(function(next) {
+			if (!func.init && !self.initialized) { return callback(new Error(utils.message(utils.messages.nonInit))); }
```
dmitrig01: a cleaner way to do this might be to just pass it in as a third parameter on the wrap function?

also -- you can move this line down to after the `var done` and just call `done(new Error)`- t thoughts

dmitrig01:
after these tweaks we're good to go - just did a full read-through

src/0_utils.js
dmitrig01: here -- `.replace(/([\?\&]_branch_match_id=\d+/, '')`

scotthasbrouck:
Alright @dmitrig01 just checking all this off, then let's release tonight!
- callback_params outside of wrap and used to call wrap
- Double checked for API-KEY and didn't see it anywhere else... am I missing it somewhere?
- updated regex for utils.cleanLinkData
- updated regex for branch_key in api.js
- affirm @return {_validator} should stay - not sure why I deleted it
- affirm (Object|null)=) not needed, removed
- correct, wrapError no longer used. removed
- agreed, just passed in function as third argument
- yep, just moved the if not init check to below done

Quick thing though, not sure what 'this' could be passed in as a third argument to wrap().

dmitrig01:
Oops sorry -- on your last point, func.init

There's a straggler in the Cordova guide -- I can fix it

scotthasbrouck:
Ah, yep that makes sense. Here's init passed in as the 3rd param to wrap(). However, do you know how define an arbitrary number of optional params for @param {function} func? line 30

dmitrig01:
i think `{function(...*): null}` lemme try it

PR #32, opened 2015-04-11
Fix for banner note checking make_new_link

scotthasbrouck:


dmitrig01:
it goes deeper than that -- I think you need it without quotes, and you need to add it to the banner options type in banner_utils

PR #38, opened 2015-04-29
Initialization states

scotthasbrouck:


scotthasbrouck:
@dmitrig01 feedback on this when you get a chance?
Also, let me know if you want me to keep looking into including the app_id in the call to referrals.

src/3_branch.js
```diff
@@ -343,10 +369,11 @@ if (CORDOVA_BUILD) { // jshint undef:false
...
  */
 Branch.prototype['setIdentity'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done, identity) {
+	var self = this;
 	this._api(resources.profile, { "identity": identity }, function(err, data) {
-		this.identity_id = data['identity_id'];
-		this.sessionLink = data['link'];
-		this.identity = data['identity'];
+		self.identity_id = data['identity_id'];
+		self.sessionLink = data['link'];
+		self.identity = data['identity'];
```
dmitrig01: ack, great catch

scotthasbrouck:
Haha, yeah.
Good to merge this?

dmitrig01:
i've now gotten halfway through writing this comment (and then getting distracted) at least twice, but:

it looks great! the one change i might make is - not sure if this belongs in the docs. for one, we don't actually expose init_state, since the minified will rename the variable to a different name. but beyond that, what's the use case for someone actually using it?

scotthasbrouck:
GREAT point. haha...
One thing that comes to mind is rather than just getting an error returned by init, perhaps they would want to stop making further SDK requests if init_state is failed? It would be easy to expose, but perhaps it's not worth adding to the amount of information already in the Readme, it is getting rather long. I'm fine not mentioning it in the docs.

dmitrig01:
yeah, i think as long as the error is easy to catch it's fine

maybe we could say something like "if init fails, all subsequent calls will also fail"

scotthasbrouck:
That makes sense. I'll make those changes now

scotthasbrouck:
Ok, should be all set

dmitrig01:
:+1: good work

PR #41, opened 2015-05-07
Integration tests

scotthasbrouck:


scotthasbrouck:
@dmitrig01 Forgot to commit and push this up the other day. Note that this is not an integration test on the minified build. The complication is that there is no way to stub out the jsonp request, unless we were to not minify the method names of _server(), and createScript(). XHR would work fine with the minified build because of Sinon's FakeXHR thing. I'll also keep brainstorming to see if I can come up with other options.

dmitrig01:
awesome! i'm pretty backed up but i may have time to review tomorrow

scotthasbrouck:
Cool no rush

dmitrig01:
hey - yknow what we shoudl do?

in all the autogenerated files which don't look like auto-genertaed files (e.g. README, this test file, the testbed/example.html), we should put a `<!-- THIS FILE IS AUTOMATICALLY GENERATED. DO NOT EDIT; EDIT xxxx INSTEAD -->`. I can already see the mistakes that will come from this :P

test/4_integration.js
```diff
@@ -1,108 +1,242 @@
-goog.require('utils');
-goog.require('Server');
-goog.require('Queue');
-goog.require('Branch');
-goog.require('resources');
 goog.require('config');
-goog.require('storage');
-
+goog.require('utils');
```
dmitrig01: do we need config and utils?

test/integration-test.html
```diff
@@ -0,0 +1,36 @@
...
+</head>
+<body>
+	<div id="mocha"></div>
+	<link rel="stylesheet" type="text/css" href="../node_modules/mocha/mocha.css" />
+	<script src="../compiler/library/closure-library-master/closure/goog/base.js"></script>
+	<script>
+		goog.addDependency("../../../../../src/0_config.js", ['config'], []);
+		goog.addDependency("../../../../../src/0_utils.js", ['utils'], []);
+		goog.addDependency("../../../../../src/0_storage.js", ['storage'], []);
+	</script>
```
dmitrig01: same question here - do we need the closure library or these dependencies?

dmitrig01:
also here - let's add to changelog

test/4_integration.js
```diff
@@ -1,108 +1,242 @@
...
-			branch.init(branch_sample_key);
-			requests[0].callback(null, browser_fingerprint_id);
-			requests[1].callback(null, { session_id: session_id, browser_fingerprint_id: browser_fingerprint_id, identity_id: identity_id });
-			requests = [];
-		}
+		it('should store in session and call open with link_identifier from hash', function(done) {
+			var assert = testUtils.plan(1, done);
+			testUtils.go("#r:12345");
+			branch.init(branch_sample_key, function(err, data) {
+				assert.equal(utils.readStore(branch._storage).click_id, '12345', 'click_id from link_identifier hash stored in session_id');
```
scotthasbrouck: Yep, I need utils here to test that the hash value was stored

dmitrig01: ahh ok cool. hmm... it's like a weird coupling of unit and integration testing. let's find other ways to check this -- for example, do a sendSMS and check the url it gets sent to

scotthasbrouck:
@dmitrig01:
- Added changelog note
- removed utils and session dependency, and just checking URLs everywhere. Note that I'm still using config, only to grab the current SDK version to check the URL for _r
- added the README notice

dmitrig01:
Hmm, I resolved the merge conflicts, but the tests don't pass for me

dmitrig01:
can you take a look?

scotthasbrouck:
all set!

dmitrig01:
sweet!

PR #42, opened 2015-05-08
Json parse data

scotthasbrouck:
Options to parse `data`, and `referring_data` in `init()` and `setIdentity()`, per https://github.com/BranchMetrics/Web-SDK/issues/40


dmitrig01:
I like it. I thought you were going to this (which I like better :P): make a new key returned, called data_parsed. Then deprecate (or don't document) the existing data key, and only document data_parsed instead.

Thoughts/

scotthasbrouck:
Oh nice! I don't recall that, but that's a much better option. I'll make that change and dump the option all together

dmitrig01:
thanks. sorry to make you duplicate work - i thought that's what we'd discussed :P

scotthasbrouck:
Np, took like 2 mins to change. I'll get the tests working

scotthasbrouck:
@dmitrig01 all set for review

test/3_branch.js
```diff
@@ -194,7 +194,7 @@ describe('Branch', function() {
 		var expectedResponse = {
 			identity_id: '12345',
 			link: 'url',
-			referring_data: { },
+			referring_data: '{ }',
```
dmitrig01: hrm, this should have been failing before, right?

scotthasbrouck: Actually, no. Because the setIdentity method didn't do anything with the data parameter previously. So it didn't matter what type it was. Not it is required to be a string, because I'm parsing it.

dmitrig01: ah, i see

dmitrig01:
two minot rhints, then we're ready to go
1. can you add to the changelog
2. can you resolve the merge conflict?

scotthasbrouck:
@dmitrig01 adds changelog and resolves merge conflict

dmitrig01:
:+1:

PR #43, opened 2015-05-08
fix markdown in src/3_branch.js

jakecadams:
@dmitrig01 not exactly sure whats going on with the cordova mods. looks like someone changed sources and maybe didn't run make?


dmitrig01:
yea. probably a new version of something or other

PR #53, opened 2015-05-18
Banner style fixes with body position

scotthasbrouck:
Fixes when banner is shown:
body { position: relative; }
body { position: absolute; }
body background image position


src/2_banner.js
```diff
@@ -154,8 +154,11 @@ banner = function(branch, options, linkData, storage) {
 			}, banner_utils.animationSpeed + banner_utils.animationDelay);
 
 			setTimeout(function() {
-				if (options.position == 'top') { document.body.style.marginTop = '0px'; }
-				else if (options.position == 'bottom') { document.body.style.marginBottom = '0px'; }
+				if (options.position == 'top') {
+					document.body.style.paddingTop = '0px';
+					document.body.style['backgroundPositionY'] = '0px';
```
dmitrig01: hmm, i think this could break something, if you have your backgroundPositionY set to somethign like 'center'... – why do we need to change it?

also - maybe we should set both marginTop and paddingTop to 0?

also - maybe we shoudl save the previous values so we can restore them when the banner is hidden? I don't want to interfere with the CSS too much though so we should have a toggle switch to turn that off if we do it... sooo many 
switches :-/

scotthasbrouck:
Ok, lot's of changes here:
1. Let's only change the inline margins, and leave paddings alone.
2. We need to save any inline margin-top value, and re-apply it with it's original unit (px, %, em, etc.), when the banner is closed. I just save this as a string, and reapply it on banner close.
3. If there is an inline margin-top, we need to convert it to pixels, and add it to the banner height of 76px, and make that the new margin-top. In every browser other than IE, this is simple by using getComputedStyle. With IE, computedStyle only returns the string in it's original units. So I wrote a banner-utils method that converts this to a raw number in pixels (basically what getComputedStyle does).
4. Yes, I don't think it's worth worrying about the background. And it would get way too messy anyway with possible values like 'center'

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ The Branch Web SDK adheres to [Semantic Versioning](http://semver.org/).
 - Specify whether the app banner is "sticky" (position: fixed) on desktop and mobile
 - Set the app banner to either never show a second time, or wait a specified number of days before showing again
 - Adds `customCSS` property to banner to allow custom styles
+### Fixed
+- Fixes body element position when banner is shown with `body { position: relative; }` or `body { position: absolute; }
```
dmitrig01: very minor but missing end `

dmitrig01:
it's all a bunch of magic to me but if you've tested it everywhere then let's do it... we should really get some banner tests going :-)

PR #54, opened 2015-05-18
Network retry on fail or timeout

scotthasbrouck:
This has been tested in IE 9, 10, and 11, Firefox, Safari, and Chrome


src/1_api.js
```diff
@@ -191,10 +194,31 @@ Server.prototype.request = function(resource, data, storage, callback) {
...
+	// If request fails, retry after X miliseconds
+	/***
+	 * @type {function(?Error,*=): ?undefined}
+	 */
+	var done = function(err, data) {
+		if (err && retries > 0) {
+			retries--;
+			window.setTimeout(function() {
+				makeRequest();
+			}, 200);
```
dmitrig01: hmm, a few things:

1. we should count 5xx errors as failures, but not 4xx errors – we may have to fiugre out a smart way to do that
2. this doesn't actually set a timeout on requests, this jsut sets a request after a certain amount of time; it would be great to actually set a timeout on requests - otherwise if we fire off an JSONP request and it never returns, nothing will happen
3. let's move the constants (retires = 2, 200) as all-caps constants at the top of the page

scotthasbrouck:
@dmitrig01 this will now only retry 5xx errors, and also timeouts.
The timeouts are actually int he jsonpRequest and XHRRequest methods themselves, set at 5000ms. I moves that value up as a constant also, to make it clearer. The 200ms setTimeout is merely a delay before trying the request again, like a retry delay, not a timeout for the request.
I'll get the tests passing in a sec, have a call with Applico

PR #56, opened 2015-05-20
Get Referring link

scotthasbrouck:


src/3_branch.js
```diff
@@ -693,8 +733,11 @@ Branch.prototype['sendSMS'] = wrap(callback_params.CALLBACK_ERR, function(done,
 		}, done);
 	}
 
-	if (utils.readKeyValue('click_id', self._storage) && !options['make_new_link']) {
-		sendSMS(utils.readKeyValue('click_id', self._storage));
+	var click_url = utils.readKeyValue('click_url', self._storage),
```
dmitrig01: let's just use getReferringLink for this, right?

src/3_branch.js
```diff
@@ -587,6 +592,41 @@ Branch.prototype['link'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
+ * );
+ * ```
+ *
+ * ___
+ *
+ * ## Sharing links via SMS
+ */
+Branch.prototype['getReferringLink'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done) {
+	var click_url = utils.readKeyValue('click_url', this._storage),
+		click_id = utils.readKeyValue('click_id', this._storage);
```
dmitrig01: hmm, i think ther eshould be a 1-1 mapping between click url and click id, right?

src/3_branch.js
dmitrig01: also in banner we should use it

scotthasbrouck: Ah yep

testbeds/web/example.html
```diff
@@ -79,7 +79,7 @@
   <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.2.0/js/bootstrap.min.js"></script>
   <script type="text/javascript">
 
-(function(b,r,a,n,c,h,_,s,d,k){if(!b[n]||!b[n]._q){for(;s<_.length;)c(h,_[s++]);d=r.createElement(a);d.async=1;d.src="https://cdn.branch.io/branch-v1.5.2.min.js";k=r.getElementsByTagName(a)[0];k.parentNode.insertBefore(d,k);b[n]=h}})(window,document,"script","branch",function(b,r){b[r]=function(){b._q.push([r,arguments])}},{_q:[],_v:1},"init data setIdentity logout track link sendSMS referrals credits redeem banner closeBanner".split(" "),0);
+(function(b,r,a,n,c,h,_,s,d,k){if(!b[n]||!b[n]._q){for(;s<_.length;)c(h,_[s++]);d=r.createElement(a);d.async=1;d.src="../../dist/web/build.js";k=r.getElementsByTagName(a)[0];k.parentNode.insertBefore(d,k);b[n]=h}})(window,document,"script","branch",function(b,r){b[r]=function(){b._q.push([r,arguments])}},{_q:[],_v:1},"init data setIdentity logout track link getReferringLink sendSMS referrals credits redeem banner closeBanner".split(" "),0);
```
dmitrig01: hmm, do we wnat to change this? or is this a built file that'll get wiped?

scotthasbrouck: Oh this get's built from onpage.js

src/3_branch.js
scotthasbrouck: Well, let's think about this. If we're calling open with a match_id, it should always return a click_url, correct? In that case, we don't need to bother saving click_id.
The only other place click_id is saved is in sendSMS, but we could instead build a URL out of it and just save it as click_url instead. Then we could entirely do away with saving click_id.

PR #59, opened 2015-05-27
Saucelabs + CircleCI

scotthasbrouck:
Saucelabs testing in 33 browser configs. If we do many more than this, testing is going to get slow. We're covering every major browser, so the only other tests we could really be doing is the same browser on multiple operating systems. I don't think that's necessary.


scotthasbrouck:
@dmitrig01 this is ready to go, but integration tests are commented out. Your call if you want to merge this now, or if I should finish that up first.

scotthasbrouck:
@dmitrig01 got integration working with sauce! Found a stupid mistake stopping it from working with Phantom, so then it worked in sauce.

scotthasbrouck:
So this is all tests, running in sauce now.

dmitrig01:
awesome!!!! :+1:

PR #60, opened 2015-05-28
Adds buildnumber grunt task

scotthasbrouck:


dmitrig01:
hmm... is there any way to just use the version number as the build number? or at least set both the version number and the build number to the same thing (using that perl regex)? just makes things simpler :-)

scotthasbrouck:
Ah, yeah, we can do that. I'm heads down in revamping the storage class. So I'll make that change later tonight.

dmitrig01:
sweet :D

dmitrig01:
ping ^

package.json-e
```diff
@@ -0,0 +1,36 @@
+{
```
dmitrig01: let's remove this file and ad to gitignore :-)

dmitrig01: then this pr is good to go

circle.yml
```diff
@@ -28,5 +28,5 @@ dependencies:
 test:
   override:
     - ./lint.sh
+    - grunt buildnumber
```
dmitrig01: this can be removed :-)

scotthasbrouck:
Cool, all set with those changes. I also removed most of the mobile tests to see if we have better luck with sauce labs timing out and giving false failures. From what I've read, SauceLabs has better desktop tests, and BrowserStack has better mobile tests because they use actual devices rather than emulators. I don't know if it's worth using both though.

PR #63, opened 2015-05-31
Storage class revamp

scotthasbrouck:
**Note**: Still needs to be manually  tested in all browsers, and not necessarily ready to be reveiwed.

This new `storage` class has the concept of temporary and permanent storage, each with several methods of storage that are all attempted with every write, as so:

**Temporary**: data is stored until the user closes their browser window.
1. sessionStorage
2. cookie with no expiration date (a session cookie)
3. `_store` object that is saved to the `branch` object.

**Permanent**: data is stored indefinitely, and survives closing the browser window.
1. localStorage
2. cookie with an expiration date 365 days in the future (will probably make this longer prior to releasing this)
3. `_store` object - not this will not persist, but it at least still lets us save data when the user has both cookies and localStorage disabled (they are browsing in private mode perhaps).

With that in mind, there are the following functions available on the storage class:
`setPermItem(key, value)`
`setTempItem(key, value`
`getItem(key)` - first will return tempItem, then permItem if no temp item is found with that key
`removeItem(key)` - will remove both temp and perm items with that key
`clear()` - clears absolutley everything
`clearTemp()` - only clears temp storage
`clearPerm()` - only clears perm storage

I've also moved all branch session functions out of `utils` and into a new `session` class. The session functions are basically a second abstraction on top of storage. I want to think through if this makes sense.


dmitrig01:
what about making a storage "interface" that is "implemented" by cookies, localStorage, and sessionStorage – it should be fairly simple to, in each of setTemp, setPerm, get, etc, specify what preference we have/what fallbacks/etc

scotthasbrouck:
@dmitrig01 check out the `src/0_storage.js` implementation now.

This includes all of the Titanium module additions, documentation, and has been tested in Cordova, Titanium, and all tests pass in web.

src/0_storage.js
```diff
@@ -1,68 +1,218 @@
...
+var BRANCH_KEY_PREFIX = 'BRANCH_WEBSDK_KEY';
+
+/** @typedef {{get:function(string), set:function(string, (string|boolean)), remove:function(string), clear:function(), isEnabled:function()}} */
+var storage;
+
+// Try catch to prevent error on unit tests
+try {
+	/** @typedef {{listProperties: function(), setString: function({string}, {string}), getString: function({string})}}*/
+	Ti.App.Properties;
+} catch(e) {}
```
dmitrig01: let's add an if titanium around this?

scotthasbrouck: Oh duh.

dmitrig01:
is this ready to go?

dmitrig01:
i think there are still a bunch of files which can be removed

dmitrig01:
lemme make a super quick pass at it

scotthasbrouck:
Cool

dmitrig01:
ack... of course now checking out the repo is super slow... ugh

dmitrig01:
please don't merge this in – I will merge with a rebase and delete the branch, so it doesn't balloon repo size

scotthasbrouck:
Sounds good

dmitrig01:
*when we're ready to merge, i will do it

dmitrig01:
in testsbeds/titanium/Branch-SDK-Testbed/modules, are those full copies? is it possible to make those aliases?

scotthasbrouck:
Got it

dmitrig01:
in `src/titanium/BranchSDK/android/libs/*/*.so` - what are these? how do they get built?

dmitrig01:
`src/titanium/BranchSDK/android/dist/branchsdk.jar` - this doesn't need to be there, right?

dmitrig01:
can we unify all the artwork? specifically - do we need to have all these different files? maybe we could just have one giant one?

if we can't do that, can we have one giant one, and have the other oenes just be symlinks to that one?

dmitrig01:
pushed my removals, if you can do those then i'll merge

dmitrig01:
also if you can mereg master in :-)

dmitrig01:
you know more about this than i do: do we need all the stuff in `src/titanium/BranchSDK/iphone/BranchSDK.xcodeproj`? or are there things we should be removing there?

scotthasbrouck:
The only reason we would keep most of these files is if we want the two testbeds apps to build cleanly.
- If we expect them to fetch the titanium modules themselves, we can delete the entire modules folder.
- The only reason you need that artwork is if you're going to submit to the store. Not important.

Also confirmed this with Rob.

So I'll make those deletions, and just expand the README to make sure people know how to build it

dmitrig01:
which modules?

dmitrig01:
I do think we want the testbeds to build cleanly

scotthasbrouck:
Ok, I can put the module back in - as in the Branch SDK module - otherwise they would just have to fetch it themselves. It will build fine without the artwork, in fact, I think titanium puts that back in there when you build the app if it isn't present.

dmitrig01:
isn't this the branch SDK? so it's referencing something two or three dirs outside of itself? can we use a symlink

dmitrig01:
?

dmitrig01:
or make a script to copy it, like we do with cordova

scotthasbrouck:
I *believe* it's the compiled Branch SDK, not the same thing in the src dir. I'm verifying this and a few other things with Rob. I'll make a push once I hear back and get it all sorted.

dmitrig01:
ahhhhhh

dmitrig01:
do you need that built code when including the branch sdk in other projects besides the testbed? how does that work?

dmitrig01:
should it be part of the build/release process?

scotthasbrouck:
I believe so, actually. If that's the case when I hear back from Rob, I'll add it to the release.sh

scotthasbrouck:
@dmitrig01 I've had Rob review this, and this seems to be as much as we can pair down the titanium testbed and still have it be easy to build. We could remove the plugin dir, but installing plugins in Appcelerator is a pain in the butt and will add a ton of steps. So I think we should just leave it for convenience since it's only a few files. The tests pass locally - so this looks good to go.

dmitrig01:
Great

scotthasbrouck:
FYI, working right now on removing most of the sauce tests so we can at least see a green checkmark here

dmitrig01:
just glanced over it-- looks perfect. want me to merge now? are you doing sauce here or anohter branch?

scotthasbrouck:
Give me a few mins on these sauce tests so I can merge that in now

scotthasbrouck:
alright @dmitrig01 let's merge this then I'll release

PR #65, opened 2015-06-03
Added 'campaign' to properties in resources.link.

elsbree:
When calling link() and sendSMS(), the 'campaign' parameter is not being sent to the server, as it's not included in the whitelisted parameters in resources.js. This PR fixes that problem.

### DIFF
--- a/src/1_resources.js
+++ b/src/1_resources.js
@@ -288,6 +288,7 @@ resources.link = {
 		"data": validator(false, validationTypes.str),
 		"tags": validator(false, validationTypes.arr),
 		"feature": validator(false, validationTypes.str),
+		"campaign": validator(false, validationTypes.str),
 		"channel": validator(false, validationTypes.str),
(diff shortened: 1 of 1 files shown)


dmitrig01:
great catch – thanks!

dmitrig01:
@scotthasbrouck can you take a look and get this to a test-passing state?

scotthasbrouck:
@dmitrig01 it's not the tests failing, it's saucelabs timing out on mobile tests. I've been going back and forth with them for a while, and they've finally responded that "this is as good as it gets" with emulated mobile tests. I'll be replacing them with something else this week. In the meantime, we could switch back to just phantomjs if we want to see passing tests on PRs.

dmitrig01:
ok - so if you run this locally it passes tests?

scotthasbrouck:
Correct, everything passes 100%

scotthasbrouck:
sweet

PR #69, opened 2015-06-09
Tweaks and fixes to mobile banner.

scotthasbrouck:
Takes care of:
- vertical centering when description flows to multiple lines
- Android green is less sickly
- Vertical centering of both android and ios close buttons
- Vertical centering of mobile action (download/open)


scotthasbrouck:
@dmitrig01

PR #70, opened 2015-06-09
Branch pub/sub observer

scotthasbrouck:
*Note yet complete, at minimum, needs tests and documentation*


scotthasbrouck:
@dmitrig01

src/3_branch.js
```diff
@@ -323,6 +338,36 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
 }, true);
 
 /**
+ * @function Branch.addListener
+ * @param {function(String)} observer - _required_ - Observing function that will recieves a string as the events.
+ * @param {String} subject - _optional_ - Optionally specify which subjects you would like to observe. If left empty, the observer will recieve all events. Currently, the only Branch subject that publishes events is the banner.
+ *
+ */
+Branch.prototype['addListener'] = wrap(callback_params.NO_CALLBACK, function(done, observer, subject) {
```
dmitrig01: wrapping would put it in the queue - we want to keep it out of the queue, so callbacks get added immediately right?

src/3_branch.js
```diff
@@ -149,6 +150,19 @@ Branch.prototype._referringLink = function() {
 	else { return null; }
 };
 
+/***
+ * @function Branch._publishEvent
+ * @param {string} event
+ * @param {string} subject
+ */
+Branch.prototype._publishEvent = function(event, subject) {
```
dmitrig01: I think most event systems just have one concept of event/subject - we can just have people bind to willCloseBanner, didCloseBanner, or both - right?

src/3_branch.js
scotthasbrouck: good call.

src/3_branch.js
scotthasbrouck: Yeah, we could just only publish events. Then they could add listeners like:
```
branch.addListener(fn); // listen for all events

branch.addListener(fn, "willCloseBanner"); // just listen for willCloseBanner events

branch.addListener(fn, ["willCloseBanner", didCloseBanner"]); // listen for 2 or more event types
```

People will probably want to listen for specific events anyway, then we can remove the concept of subjects. Sound good?

scotthasbrouck: Then, _publish event would just become:
```
Branch.prototype._publishEvent = function(event)...
```

dmitrig01: yea - we can just do one at a time for now:
```
branch.addListener("willCloseBanner", fn);
branch.addListener("didCloseBanner", fn);
```
also - the typical syntax is to put the event first, the function second.

scotthasbrouck: Cool. I'll make both those changes

scotthasbrouck:
@dmitrig01 all set with those changes. Not sure why it's saying there are merge conflicts, I've already merged from master and resolved everything. I'll just merge locally if need be

src/2_banner.js
```diff
@@ -170,8 +179,10 @@ banner = function(branch, options, linkData, storage) {
 		setTimeout(function() {
 			if (options.position == 'top') { element.style.top = '0'; }
 			else if (options.position == 'bottom') { element.style.bottom = '0'; }
+			branch._publishEvent("didShowBanner", "banner");
 		}, banner_utils.animationDelay);
 
 		return closeBanner;
 	}
+	else { branch._publishEvent("willNotShowBanner", "banner"); }
```
dmitrig01: publishEvent deson't take a second argument

dmitrig01:
great thanks!!

PR #72, opened 2015-06-09
Calebrash patch 1

scotthasbrouck:


scotthasbrouck:
@calebrash We actually generate WEB_GUIDE.md from src/3_branch.js with JSDoc. I made your change to the template and regenerated it. Thanks for the fix!

PR #76, opened 2015-06-25
fix parse error

scotthasbrouck:
Even inside a try/catch block, closure compiler still throws an error if `value` is a string.


scotthasbrouck:
@dmitrig01 @derrickstaten this fixes the open issue

PR #77, opened 2015-06-30
Fixes to sauce tests

scotthasbrouck:
Not quite ready for review, I want to double check this tomorrow.


scotthasbrouck:
@dmitrig01 this should be good to go. Figured out that tests on most mobile devices were failing silently. Also figured out a few more problems that were leading to nearly all of the timeouts. So we should have much more reliable green checks with tests being run on all the environments

dmitrig01:
Cool!

Issue #78, opened 2015-07-01
Cordova plugin - handleOpenUrl is never called


wedgybo:
I'm having a lot of consistency problems with the cordova integration. It looks like BNCDevice.m pretty much does nothing from a standard install using 

cordova plugin add https://github.com/BranchMetrics/Web-SDK.git

The handleOpenURL function is never called because the plugin is never told to listen for the handleOpen notification

The getOpenData seems redundant also. I can see some calls to it inside the build.js file supplied with the plugin, but the breakpoint in either getInstallData or getOpenData is never hit. Again this is in minified code so it's tricky to see what's going on but I'll start digging deeper to see why.

I guess what I'd like to know is are these things meant to be called. Or is the plugin actually redundant now and it should all work from the WebSDK? I'm getting a bit confused and I don't want to waste time going down the wrong path.

Any advice would be appreciated!

aaustin:
Hey @wedgybo - this is a pretty atrocious list of faults that we can dig into right away. I'm pretty sure you don't need to use handleOpenURL, as all you need to do is to execute this code once the ‘deviceready’ event fires and each time the ‘resume’ event fires

  branch.init("YOUR BRANCH KEY HERE", function(err, data) {
    app.initComplete(err, data);
  });

You need to use the Cordova version instead of the Web SDK version because it does execute some device specific code, but the methods are the same as the Cordova library.

It seems like you're on the right path!

wedgybo:
Ok. I'm calling the branch.init function on both of those events. Here's my test code.

```
function initBranch() {
    console.log('BRANCH INIT');
    branch.init(BRANCH_KEY, { isReferrable: true }, function (err, data) {
      console.log('BRANCH DATA', err, data);
    });
}

document.addEventListener('deviceready', initBranch, false);
document.addEventListener('resume', initBranch, false);
```

Now I see BRANCH INIT and  BRANCH DATA on my cold start no problem, from a warm start it only runs BRANCH INIT.

The problem appears to lie somewhere here. https://github.com/BranchMetrics/Web-SDK/blob/master/src/3_branch.js#L306

The SDK know's there's a session so it fires on with the web flow. Unfortunately it doesn't have the information from the click so the data object is never updated and always returns the first data object found.

[...]

PR #79, opened 2015-07-02
Fixes issue with Cordova SDK not loading previous session

scotthasbrouck:
Simple fix for Cordova and Titanium. They should not be calling *just* finishInit if a previous session is present.

Addresses issue #78


dmitrig01:
good catch

scotthasbrouck:
Credit goes to: @wedgybo

wedgybo:
Thanks guys I'll check this out and see if it fixes my issue!

Some more information though. I think the issue is actually with the close statement being called on pause.

It's being added to the queue but in my case it's never actually executed. I've not quite worked out how the queue operates yet. But my guess is close() needs to jump to the front of the queue, or the queue needs cleared before it's added as it looks like it runs next() once then exits.

I think the correct fix is to make sure close() is actually executed as this will clear out the session id and that check would fail as expected.

Just my thoughts! I'll report back shortly with an updated on how this PR works for me!

wedgybo:
I'm afraid it wasn't enough so I got stuck in properly. I've raised PR #80 to have a discussion about it.

aaustin:
so can we merge this one?

wedgybo:
I don't think it will hurt, but it doesn't fix the issue I was reporting.

Happy 'Merica day guys! Enjoy your holiday. We're having a mini celebration with you here in rainy Scotland :us: :gb: We can sort out the proper fixes next week!

scotthasbrouck:
hey @wedgybo thanks for the happy 'Merica day wishes haha
I found the root cause to the 2nd issue over the weekend. All of our API endpoints are stored in src/1_resources.js. Some of the endpoints are used either by only the web SDK, or only the Cordova SDK. The end point for the close method somehow ended up in the Web SDK block, moving it to the Cordova block fixed the issue. I did a lot of cleanup in the testbed app for future engineers wanting to work with Cordova, and wrote a suite of integration tests for Cordova - previously there were no tests. So moving forward this should help prevent obvious bugs like this!
This branch was misspelled and it annoyed me :) So I finished this fix on a new branch called cordova_fix, which I'll make a PR for.

PR #80, opened 2015-07-03
Fixes issues with closing the branch session in cordova

wedgybo:
This PR fixes the issue with close failing and the queue not processing things properly.

### DIFF
--- a/src/0_queue.js
+++ b/src/0_queue.js
@@ -15,6 +15,6 @@ var Queue = function() {
 	};
 	return function(task) {
 		queue.push(task);
-		if (queue.length == 1) { next(); }
+		next();
 	};
 };
--- a/src/0_storage.js
+++ b/src/0_storage.js
@@ -10,7 +10,7 @@ goog.require('utils');
 
 var COOKIE_DAYS = 365;
 
-var BRANCH_KEY_PREFIX = 'BRANCH_WEBSDK_KEY';
+var BRANCH_KEY_PREFIX = 'branch_';
 
 /** @typedef {undefined|{get:function(string), set:function(string, (string|boolean)), remove:function(string), clear:function(), isEnabled:function()}} */
 var storage;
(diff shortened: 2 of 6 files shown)


src/0_storage.js
```diff
@@ -50,16 +50,28 @@ var webStorage = function(perm) {
...
+					allKeyValues[trimPrefix(key)] = retrieveValue(storageMethod.getItem(key));
+				}
 			}
 			return allKeyValues;
 		},
 		get: function(key) { return retrieveValue(storageMethod.getItem(prefix(key))); },
 		set: function(key, value) { storageMethod.setItem(prefix(key), value); },
 		remove: function(key) { storageMethod.removeItem(prefix(key)); },
-		clear: function() { storageMethod.clear(); },
+		clear: function() {
```
wedgybo: The clear function for local storage was wiping out everything so I've made it work based on the key prefix

dmitrig01:
hey @wedgybo - this looks great - thanks so much. a few minor comments:

src/0_storage.js
```diff
@@ -10,7 +10,7 @@ goog.require('utils');
 
 var COOKIE_DAYS = 365;
 
-var BRANCH_KEY_PREFIX = 'BRANCH_WEBSDK_KEY';
```
dmitrig01: let's keep this - otherwise it'll break old integrations

src/3_branch.js
```diff
@@ -285,6 +285,7 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
 			data = setBranchValues(data);
 			if (CORDOVA_BUILD || TITANIUM_BUILD) { // jshint undef:false
 				var first = self._storage.getAll();
+				// I feel this check is wrong?
```
wedgybo: The first check still doesn't work as expected. I think there's something up with the logic here.

src/0_storage.js
dmitrig01: great catch - don't we want to test `item.indexOf(BRANCH_KEY_PREFIX) === 0`?

src/0_storage.js
wedgybo: Cool. This is the area I can help fix without you guys.

Basically I think you're trying to store two separate keys. One for the first time a session is detected. `branch_data` and the regular `branch_session` so I'm struggling to move forward without understanding what you guys were intending to check for the first session stored in cordova

src/0_storage.js
wedgybo: Yes, but it's set to something that doesn't match 'branch_session' which is the only key you want to clear when closing the session

src/0_queue.js
```diff
@@ -15,6 +15,6 @@ var Queue = function() {
 	};
 	return function(task) {
 		queue.push(task);
-		if (queue.length == 1) { next(); }
```
wedgybo: I don't think this is right solution, but it got it working properly for now.

It looks like the queue never clears, so this condition would stop the close() from running

dmitrig01: hmm, i bet there's a case where the callback isn't getting called...

PR #83, opened 2015-07-07
stupid ie10/11 bug with order of XHR properties

scotthasbrouck:
Confirmed on my windows machine, this fixes the issue with IE 10/11. This was the same bug we had before, and it snuck back in.


aaustin:
Do we need to delete it on Cordova and Titanium too? Is it extraneous code?

scotthasbrouck:
It wouldn't affect titanium or cordova - this is only an IE bug. I remember confirming that in gory detail a few months ago, plus deleting it here will delete it in all of the build targets. Web, Titanium, and cordova

derrickstaten:
Can we get this out for the partner as soon as CircleCI is back up?

dmitrig01:
@aaustin those are compiled files

PR #84, opened 2015-07-07
prevents embedding or initializing branch twice

scotthasbrouck:
@dmitrig01 all tests pass. The prevents:
Embedding branch more than once
Initializing Branch more than once


src/onpage.js
```diff
@@ -8,7 +8,7 @@
  */
 
 (function(root, doc, scriptStr, branchStr, createCallback, branchSdk, funcs, i, scriptTag, firstScript) {
-	if (!root[branchStr] || !root[branchStr]._q) {
+	if (!root[branchStr] || (root[branchStr] && !root[branchStr]._q)) {
```
dmitrig01: this will do exactly the same thing.

but i think the problem is, `root[branchStr]._q` is truthy when there is a branch sdk on the page AND it hasn't loaded the full script, only this shim.

what we really want is to see if any branch sdk is loaded at all. so maybe either remove the second part or make it `!root[branchStr] || root[branchStr][funcs[i]]` (to veriyf that what's on the page is actually the branch sdk)

src/onpage.js
```diff
@@ -18,9 +19,8 @@
 		scriptTag.src = 'SCRIPT_URL_HERE';
 		firstScript = doc.getElementsByTagName(scriptStr)[0];
 		firstScript.parentNode.insertBefore(scriptTag, firstScript);
-
-		root[branchStr] = branchSdk;
 	}
+	else { console.log("Branch is already embedded" ); }
```
dmitrig01: honestly i think it's fine if we fail silently, since they'll just be execuitng functions on an already-defined brnach object

scotthasbrouck: makes sense

dmitrig01: maybe we can verify it's branch with like `else if (!(typeof root[branchStr] == 'object' && root[branchStr].init && root[branchStr].data && root[branchStr].link)) { console.log('Conflict: you already have another global varibale named branch') }`

Kirkkt:
@scotthasbrouck 
I will close this PR in 7 days because it's a bit outdated. Let me know if you think otherwise.

dmitrig01:
@Kirkkt this one is also relevant - can you pick it up?

aaustin:
This doesn't seem like it's solving a our partners have. Closing.

PR #85, opened 2015-07-15
Fix issue with subsequent link clicks not updating link click data

scotthasbrouck:


scotthasbrouck:
Figured out the latest issue with sauce. As long as we don't run it too many times, back to back, I think we'll get green checks now

scotthasbrouck:
@dmitrig01

src/3_branch.js
```diff
@@ -294,13 +298,12 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
 		done(err, data && utils.whiteListSessionData(data));
 	};
 
-	if (WEB_BUILD && sessionData  && sessionData['session_id']) {
-		finishInit(null, sessionData, false);
+	// Pick up here
+	if (WEB_BUILD && sessionData  && sessionData['session_id'] && (link_identifier === sessionData['sessionData'] || link_identifier === sessionData['click_id'])) {
```
dmitrig01: typo

scotthasbrouck:
@derrickstaten this works for me. Can you verify that if a session is already loaded in the web browser, and the link click id matches, the Web SDK will not call open and simply load the link click data from the session.
And also, if the link click id does *not* match what is in the session, that the Web SDK will call open and load the new link click data.

Tests are included in this to make sure we don't break it in the future
Then we can get this out!

derrickstaten:
:+1:

dmitrig01:
@scotthasbrouck did we add the localStorage?

if so let's do it

scotthasbrouck:
@dmitrig01 Yes, this saves the first session to localStorage now

dmitrig01:
great. :+1:

PR #86, opened 2015-07-17
logout

dmitrig01:
@scotthasbrouck


src/3_branch.js
```diff
@@ -464,7 +466,27 @@ Branch.prototype['setIdentity'] = wrap(callback_params.CALLBACK_ERR_DATA, functi
  */
 /*** +TOC_ITEM #logoutcallback &.logout()& ^ALL ***/
 Branch.prototype['logout'] = wrap(callback_params.CALLBACK_ERR, function(done) {
-	this._api(resources.logout, { }, done);
+	var self = this;
+	this._api(resources.logout, { }, function(err, data) {
+		if (err) { done(err); }
+
+		var data = {
```
scotthasbrouck: couldn't we just reassign like `data = {...`
don't think we need to declare it since `data` is an argument in the callback

scotthasbrouck:
This looks good, and works for me. I'll write some tests on this branch

src/3_branch.js
dmitrig01: oops yeah. we should probably actually name this variable something else then though. or just pass it in as a literal into session.set

scotthasbrouck:
@dmitrig01 tested this manually, and wrote unit and integration tests and it looks good

dmitrig01:
i'm guessing these failures are just sauce flukes?

scotthasbrouck:
Yep, it's just sauce yet again.
All the tests pass locally

PR #89, opened 2015-07-22
has_app updated when session already present

scotthasbrouck:


scotthasbrouck:
@derrickstaten @dmitrig01 feel free to check this out

Scenarios tested:
1. You Don’t have a session, which means has_app will not be called (check this in the inspector -> Network with requests filtered to just XHR, only a call to `open` should be made)
2. You do have a session, which means it keeps that session, but calls has_app, and just updates that one property of the session (only a call to `has_app` should be made.
3. You have a session, you switch tabs o rminimize your browser then open the branch sdk tab again, this should trigger a call to `has_app` and updates the session
4. You have a session and the smart banner open, you have not installed the app, so it says 'Download' or whatever. You switch tabs, minimize the window etc, and install the app. Then you come back to the web sdk. A call to `has_app` should be made, and if has_app changes from false to true, the banner text should update to 'Open'.

derrickstaten:
shit we need to get this live

derrickstaten:
We need to get this live. Sorry I haven't reviewed. I don't think I'll get to this before the flight

derrickstaten:
I can try to review on the plane

CHANGELOG.md
```diff
@@ -3,7 +3,15 @@ All notable changes to the Branch Web SDK will be documented here.
...
 ## [VERSION] - unreleased
+<<<<<<< HEAD
+### New Functionality
+- The `has_app` property in the session, and the mobile device action text automatically updates when a user installs the app - without reloading the page!
+### Improvements
+=======
+
+## [v1.6.5] - 2015-07-28
 - Fix error that prevented subsequent APi calls after logging out on Cordova and Titanium
+>>>>>>> master
```
dmitrig01: :'(

src/3_branch.js
```diff
@@ -301,8 +310,28 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
 		done(err, data && utils.whiteListSessionData(data));
 	};
 
+	var allTheWaysToSayHidden =
```
dmitrig01: lol love it

src/3_branch.js
```diff
@@ -301,8 +310,28 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
+		[{ 'hidden': 'visibilitychange' },
+		{ 'mozHidden': 'mozvisibilitychange' },
+		{ 'msHidden': 'msvisibilitychange' },
+		{ 'webkitHidden': 'webkitvisibilitychange' }],
+
+	attachVisibilityEvent = function(data) {
+		for(var index = 0; index < allTheWaysToSayHidden.length; index++) {
+				var hidden = Object.keys(allTheWaysToSayHidden[index])[0],
```
dmitrig01: what about just making the allTheWaysToSayHidden an array, so like
```
// be sure ot add a comment here explaining what these are
var allTheWaysToSayHidden = [['hidden', 'visibilitychange'],['mozHidden', 'mozvisibilitychange'], ...;
```
that way you prevent the nasty Object.keys

alternately tbh since there are only four of them, it might be easiest to just spell out each one

src/3_branch.js
```diff
@@ -301,8 +310,28 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
+				event = allTheWaysToSayHidden[index][hidden];
+			if (typeof document[hidden] !== 'undefined') {
+				document['addEventListener'](event, function() {
+					if (!document[hidden]) { check_has_app(data, finishInit); }
```
dmitrig01: this isn't going to work, hidden is in the wrong closure so it won't get captured properly... need an anoymous funciton :-( (alternately you could just spell out each visibility event separately)

src/3_branch.js
```diff
@@ -301,8 +310,28 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
+				}, false);
+				break;
+			}
+		}
+	};
+
 	if (WEB_BUILD && sessionData  && sessionData['session_id'] && (utils.processReferringLink(link_identifier) === sessionData['referring_link'] || link_identifier === sessionData['click_id'])) {
-		finishInit(null, sessionData);
+		attachVisibilityEvent(sessionData);
+		check_has_app(sessionData, finishInit);
```
dmitrig01: I believe we also attach the has_app state to the init call, so we shouldn't have to make another call on init

src/3_branch.js
```diff
@@ -286,6 +286,15 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
 	link_identifier = WEB_BUILD ? (utils.getParamValue('_branch_match_id') || utils.hashValue('r')) : (url ? utils.getParamValue(url) : null),
 	freshInstall = !sessionData || !sessionData['identity_id'],
 
+	check_has_app = function(data, cb) {
+		if (!sessionData) { sessionData = session.get(self._storage); }
```
dmitrig01: I would be hesitant of using this outer scope's sessionData, since another function could have changed it in the mean time. Let's do the sesion.get every time, and use a `var` so it gets scoped only to this function

src/3_branch.js
```diff
@@ -286,6 +286,15 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
+		self._api(resources.hasApp, { "browser_fingerprint_id": sessionData['browser_fingerprint_id'] }, function(err, has_app) {
+			if (has_app != data['has_app'] && has_app) { self._publishEvent("downloadedApp"); }
+			data['has_app'] = has_app;
```
dmitrig01: maybe we should just set the session directly?

dmitrig01: we also need some way of notifying the banner that it was changed

src/3_branch.js
dmitrig01: are you sure we want to call finishInit again?

dmitrig01: feels like that could lead to unintended consequences

src/3_branch.js
dmitrig01: oh never mind, you do do that, very nice!

docs/web/3_branch_web.md
```diff
@@ -740,6 +740,7 @@ branch.addListener(listener);
 - willSendBannerSMS
 - sendBannerSMSError
 - didSendBannerSMS
+- downloadedApp
```
dmitrig01: maybe we should document what these are?

test/3_branch.js
```diff
@@ -214,6 +214,29 @@ describe('Branch', function() {
...
+			branch.init(branch_sample_key);
+			requests[0].callback(null, browser_fingerprint_id);
+			requests[1].callback(null, { session_id: session_id, browser_fingerprint_id: browser_fingerprint_id, identity_id: identity_id });
+
+			requests = [ ];
+			branch = initBranch(false, true), assert = testUtils.plan(2, done);
+			branch.init(branch_sample_key);
+			assert.equal(requests.length, 1, 'one requests made');
+			assert.deepEqual(requests[0].resource.endpoint, "/v1/has-app", "Request to has_app made");
+		});
```
dmitrig01: can we try triggering a hide/show and seeing if the call is made? also if the event is fired?

src/3_branch.js
scotthasbrouck: This is actually the case where we don't call init, because we already have a session, so we do want to call it here. Where we do call init at L#381, I'm just calling attachVisibilityEvent.

src/3_branch.js
scotthasbrouck: Yeah, probably a good point - let's not call finishInit again. Even if it doesn't have unintended side effects now, someone down the road may add something that could have unintended side effects thinking the function is only called when init is done.

src/3_branch.js
scotthasbrouck: agreed, let's set the session directly

src/3_branch.js
```diff
@@ -299,14 +311,37 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
+		}
+		else if (typeof document.mozHidden != undefined) {
+			hidden = "mozHidden",
+			changeEvent = "mozvisibilitychange";
+		}
+		else if (typeof document.msHidden != undefined) {
+			hidden = "msHidden",
+			changeEvent = "msvisibilitychange";
+		}
+		else if (typeof document.webkitvisibilitychange != undefined) {
```
dmitrig01: i think you mean document.webkitHIdden?

scotthasbrouck: Ah good catch. The 1st case of document.hidden probably worked since I have an up to dat safari

aaustin:
sorry to poke on this, but could we try to get this merged ASAP? @scotthasbrouck or @derrickstaten

src/1_api.js
```diff
@@ -67,29 +87,18 @@ Server.prototype.getUrl = function(resource, data) {
 		if (resource.params.hasOwnProperty(k)) {
```
Kirkkt: indent is off

src/1_api.js
```diff
@@ -51,12 +51,32 @@ Server.prototype.serializeObject = function(obj, prefix) {
  * @param {Object.<string, *>} data
  */
 Server.prototype.getUrl = function(resource, data) {
-	var k, v, err;
-	var url = resource.destination + resource.endpoint;
+	var k, v, err,
```
Kirkkt: we don't really need to declare k here, do we? Instead we can restrict its scope to the loops only.

Similarly, v can be downgraded to a local var on line 90

src/1_api.js
```diff
@@ -51,12 +51,32 @@ Server.prototype.serializeObject = function(obj, prefix) {
...
+			if (typeof destinationObject == 'undefined') { destinationObject = { }; }
+			if (data['branch_key'] && branch_key.test(data['branch_key'])) {
+				destinationObject['branch_key'] = data['branch_key'];
+			}
+			else if (data['app_id'] && branch_id.test(data['app_id'])) {
+				destinationObject['app_id'] = data['app_id'];
+			}
+			else {
+				// way to return this!!
+				err = utils.message(utils.messages.missingParam, [ resource.endpoint, 'branch_key or app_id' ]);
```
Kirkkt: why not `destinationObject['err'] = ...`?

src/1_api.js
```diff
@@ -149,9 +158,11 @@ Server.prototype.XHRRequest = function(url, data, method, storage, callback) {
 		req.onerror = function(e) {
 			if (req.status === 402) {
 				callback(new Error('Not enough credits to redeem.'), null, req.status);
```
Kirkkt: nit: assign this message to a string to avoid copy-pasta

src/3_branch.js
```diff
@@ -286,6 +286,18 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
 	link_identifier = WEB_BUILD ? (utils.getParamValue('_branch_match_id') || utils.hashValue('r')) : (url ? utils.getParamValue(url) : null),
 	freshInstall = !sessionData || !sessionData['identity_id'],
 
+	chackHasApp = function(sessionData, cb) {
```
Kirkkt: chack -> check?

src/1_banner_html.js
```diff
@@ -108,9 +117,9 @@ banner_html.div = function(options, action) {
  * @param {banner_utils.options} options
  * @param {storage} storage
  */
-banner_html.markup = function(options, storage) {
+banner_html.markup = function(options, storage, branch, doc) {
```
Kirkkt: you are not using `doc`, also, care to update the jsdoc for branch?

src/3_branch.js
```diff
@@ -286,6 +286,18 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
+		var currentSessionData = sessionData || session.get(self._storage);
+		self._api(resources.hasApp, { "browser_fingerprint_id": currentSessionData['browser_fingerprint_id'] }, function(err, has_app) {
+			if (has_app != currentSessionData['has_app'] && has_app) {
```
Kirkkt: can we simplify this to below since it's just a boolean?

```
if (has_app && !cSD['h_a']) {
    cSD['h_a'] = true;
}

File: src/3_branch.js [hunk 18]
```diff
@@ -286,6 +286,18 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
+				currentSessionData['has_app'] = has_app;
+				session.update(self._storage, currentSessionData);
+				self._publishEvent("downloadedApp");
+			}
```
Kirkkt: wild thought: should we also log for "didNotDownloadApp"?

src/3_branch.js
```diff
@@ -299,14 +311,37 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
 		// Keep android titanium from calling close
 		if (self.keepAlive) { setTimeout(function() { self.keepAlive = false; }, 2000); }
 		done(err, data && utils.whiteListSessionData(data));
+	},
+
+	attachVisibilityEvent = function() {
+		var hidden, changeEvent;
+		if (typeof document.hidden != undefined) {
```
Kirkkt: do you mean `'undefined'` instead?

src/3_branch.js
```diff
@@ -299,14 +311,37 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
+			hidden = "hidden",
+			changeEvent = "visibilitychange";
```
Kirkkt: why all lower case?

src/3_branch.js
```diff
@@ -266,11 +266,11 @@ Branch.prototype['init'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
-			data['referring_link'] = utils.processReferringLink(data['referring_link']); //.substring(0, 4) != 'http' ? 'https://bnc.lt' + data['referring_link'] : data['referring_link'];
+			data['referring_link'] = utils.processReferringLink(data['referring_link']);
 		}
 
 		if (!data['click_id'] && data['referring_link']) {
-			data['click_id'] = utils.clickIdFromLink(data['referring_link']); // .substring(data['referring_link'].lastIndexOf('/') + 1, data['referring_link'].length);
+			data['click_id'] = utils.clickIdFromLink(data['referring_link']);
 		}
 
 		if (CORDOVA_BUILD || TITANIUM_BUILD) { // jshint undef:false
```
Kirkkt: fwiw, I believe rather than doing this jshint comment you can just `/*globals C_B T_B*/` above

src/1_api.js
scotthasbrouck: Goo call - honestly, the entire getUrl method could probably be refactored at some point. I haven't done much with it since I took this over.

Kirkkt: yep, I agree. I can do the refactory later on in a separate PR. In the meantime, allow me to leave this comment here kind of like a todo for myself. Let me know if you think otherwise :)

scotthasbrouck: Sounds good. I just started on it with some of the easy changes - but I can wrap that up and just push up here. Then you can take it from there.

Oh a tip, it you do any refactoring to the `XHRRequest` method - which would be another good thing to do, be very careful of the order that you set properties on the request when it comes to Internet explorer. For example, if you set `req.timeout` *before* calling `req.open`, you'll have an indecipherable error in IE 9 and 10.

We have standard error messages in `0_utils.js`, those XHR errors could certainly go in there.

src/3_branch.js
scotthasbrouck: Check this out: https://developer.mozilla.org/en-US/docs/Web/Guide/User_experience/Using_the_Page_Visibility_API

I guess because the internet lords didn't like camel or snake case.

src/3_branch.js
scotthasbrouck: ha, i did.

src/3_branch.js
Kirkkt: will address in a separate PR

src/1_api.js
Kirkkt: will address in a separate PR

dmitrig01: to be fair - in javascript all var declarations are hoisted to the function scope, there is no block scope - so it doesn't really matter. just a question of keeping code as readable as possible

Kirkkt: Sweet! I didn't know this. Thanks a ton!

src/1_api.js
dmitrig01: what about throwing the error here?

src/1_api.js
dmitrig01: (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting)

Kirkkt:
PTAL

WEB_GUIDE.md
```diff
@@ -844,14 +844,15 @@ branch.addListener(listener);
...
-- didSendBannerSMS
+- *willShowBanner*: `banner()` called, and the smart banner is about to be shown.
+- *willNotShowBanner*: `banner()` called, and the smart banner will not be shown. No more events will be emitted.
+- *didShowBanner*: Smart banner animation started and was is being shown to the user.
+- *willCloseBanner*: `closeBanner()` calle, and the smart banner will close.
+- *didCloseBanner*: Smart banner close animation started, and is closing.
+- *willSendBannerSMS*: Phone number in correct format, and will attempt to send SMS.
+- *sendBannerSMSError*: `sendSMS()` error returned.
+- *didSendBannerSMS*: SMS successfully sent.
+- *downloadedApp*: User installed app, and banner text updated.
```
Kirkkt: rename to `didDownloadApp`

derrickstaten:
Evaluating this PR not on the basis of "does it properly make a server request that gets back `has_app: true` when it should", but rather "if has_app is true, do we store it and modify the banner correctly," which @Kirkkt says it does. I'll be testing other things to make sure we didn't break anything.

derrickstaten:
We'll need a second investigation / PR to fix this end to end.

derrickstaten:
:+1: 

Note that the button text still will not update because there's another issue (to be identified), potentially on the backend. But this branch--its scope being `if has_app is true, do we store it and modify the banner correctly`--is solid

Kirkkt:
[UPDATE] Talked to @dmitrig01 offline: browsers may discard third party cookies, when a website embedded web SDK sends a request to branch.init() the `browser_fingerprint_id` is returned and stored in the local cookies, and since it originates from bnc.lt, browsers may treat it as third-party cookie. But when the use clicks "download" button from the banner, the visit is towards bnc.lt, which makes the cookie first party instead of third party. When we check for has_app using the `browser_fingerprint_id` contained in the said cookie, it introduces server-client inconsistency.

The proper fix will be to query bnc.lt/_r for every checkHasApp SDK function call to obtain the first cookie, and store it in the current session data.

Fix coming,

derrickstaten:
This is so exciting

dmitrig01:
:+1:

PR #92, opened 2015-07-27
fix not saving device_fingerprint on logout

scotthasbrouck:
Really simple fix for Cordova (also woudl affect Titanium)

Bug: Upon logging out, the session looses the `device_fingerprint_id`, which would cause errors for:
1. all subsequent API requests, because the required field `device_fingerprint_id` is missing, and
2. Relaunching the app and calling open, because `device_fingerprint_id` is missing

So the new procedure for fixing this on Cordova and Titanium will be to merge the PR here, int eh Web SDK. Then, updating the submodule, and running make in each of the other SDKs will update build.js in each project.


derrickstaten:
seems to work for me :+1: 

restarting tests

PR #95, opened 2015-08-03
Kirk

Kirkkt:
@jakecadams @derrickstaten 
save one layer of indent for the if statement

This if statement is a monster. A quick way to save one layer of indentation is to negate it and return early.

Also (partially) fixed hard tabs. Will fix universally later. refer to https://github.com/BranchMetrics/Android-Deferred-Deep-Linking-SDK/pull/108


jakecadams:
@Kirkkt code looks good. Would be awesome if you could look into see why all the tests aren't passing. 
We use circleci for continuous integration. You can log on with github and have access to all the our repo deploys. here is the entry point for tests from the circle.yml file: https://github.com/BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK/blob/master/circle.yml#L31

PR #98, opened 2015-08-05
use spaces for tabs for js and html files

Kirkkt:
@jakecadams 

Trivial change - only modified whitespace chars


jakecadams:
@Kirkkt so actually use tabs in js and html across our code base. I know their is the age old debate on tabs v. spaces, but IMO it really doesn't matter, but once you do pick one, you need to stick with it. So all the above files were passing our style guide. So for sanity sake, lets keep the repo in tabs.

PR #99, opened 2015-08-05
Kirk fixlint

Kirkkt:
@jakecadams @chanian 

Fixed trivial lint.sh errors, spacing, strong equal checks, etc.

This is another small one, basically no biz logic has been changed (except for === for ==, etc, but they are still simple). I left all the "variable abc is undefined" errors for a later PR.

cc: @dmitrig01 I gave you a todo FYI


src/0_storage.js
```diff
@@ -17,7 +19,8 @@ var storage;
 
 if (TITANIUM_BUILD) {
 	/** @typedef {{listProperties: function(), setString: function({string}, {string}), getString: function({string})}}*/
-	Ti.App.Properties;
+	// todo(dmitri): what does this line do? do you actuall mean `prop = Ti.App.Properties;`?
```
chanian: This looks like a no-opt, but would prefer keeping todo's out of the public SDK. Can you a separate issue to get clarity on this. Can probably confirm with Scott today too.

chanian:
:+1: other than the small comment, it looks good to me.

src/0_storage.js
jakecadams: Yeah you can always use https://github.com/BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK/issues for that, and tag the person and have a reference to the code you're talking about. (@Kirkkt if you click on the original line number, it will give you an anchored url to that line in that file i.e. https://github.com/BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK/pull/99/files#diff-6fafd7c1be3ed1a7821a1e31716d7fd0L20)

Kirkkt: Great comments. Left it intact instead.

jakecadams:
IMO, I would take a lot of care with refactoring other developer's choices on strong / weak equality. We know that in most cases, deep equality is better design, but we need to be positive that their use of weak equality wasn't intentional. In the above choices we know that the `indexOf` api will always return -1 to N, and were comparing it to a explicit 0, so were safe either way. But, if we had `foo == bar`, things get a lot more hairy.

So in summary. Use === to your hearts content when writing orignal code, just be cautious when refactoring :)

jakecadams:
great thanks for doing this! :+1:

PR #109, opened 2015-08-11
Fix typo in the doc

Nihisil:


Kirkkt:
Thank you so much!

Kirkkt:
@Nihisil 
FYI, I didn't know this until just now - this segment of the doc (not sure if the entire doc) is autogenerated from https://github.com/BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK/blob/master/src/3_branch.js#L748

I will send out a permanent fix by changing that blob of comments shortly.

PR #118, opened 2015-08-13
expose data.data_parsed alongside with data.data

Kirkkt:
data.data_parsed is a JSON object while data.data is a string.

We have been providing data.data in the past but it makes developers life much easier to have a well formatted JSON object.

Also added tests.


derrickstaten:
:+1:

PR #125, opened 2015-08-15
Kirk lint

Kirkkt:
massive fix on jshint and common js anti-patterns


src/4_initialization.js
```diff
@@ -35,7 +35,7 @@ else if (typeof exports === 'object') {
 if (!TITANIUM_BUILD) {
 	// Always make a global.
 	if (window) {
-		window['branch'] = branch_instance;
+		window.branch = branch_instance;
```
dmitrig01: most of the places we do this are actually intentional: see "Using string names to refer to object properties" in https://developers.google.com/closure/compiler/docs/limitations#implications-of-global-variable-function-and-property-renaming

dmitrig01: (this is present in many places - it's possible we have to change the linting rules to not error on this)

Kirkkt: Our `.jscsrc` and `.jshintrc` indeed need updates. Good point! Subsequential changes appending...

PR #129, opened 2015-08-18
NOP

Kirkkt:


Kirkkt:
@rpetit 

I will delete this branch in 7 days because it's rather outdated. Let me know if you think otherwise.

dmitrig01:
i think this was actually merged in

PR #132, opened 2015-08-19
NOP

Kirkkt:
@scotthasbrouck 
I will delete this branch in 7 days because it's a bit outdated. Let me know if you think otherwise.


dmitrig01:
i think we already merged this in (with a rebase though), just forgot to delete the brnach

PR #138, opened 2015-08-21
fix broken cordova links

Kirkkt:
@derrickstaten @dmitrig01 

PTaL (I also introduced a few trivial line breaks just to make my text editor sane)


dmitrig01:
thanks for slowly cleaning this mes

PR #143, opened 2015-08-24
Fix referring_link doesn't get passed into sendSMS

Kirkkt:
@dmitrig01 @jakecadams 

This happens when the option `make_new_link` is off (as default). retrieving referring_link returned null, but it should be a valid c link instead.
The reason is we were retrieving `referring_link` the wrong way. This is a quick fix, in the future I need to consolidate the setter getter for all sessionData fields


dmitrig01:
this doesn't look like it's actually changing anything - what are you intending to do?

dmitrig01:
:+1:

PR #144, opened 2015-08-24
metadata should be stringified as one object

derrickstaten:
@Kirkkt @dmitrig01 metadata should be stringified as a single object. on the api service, only two endpoints use metadata, `/v1/event` and `/v1/eventresponse`. I can't find `eventresponse` in the web-sdk, so I feel confident that I'm not breaking anything else in the web sdk by using it here.

One partner is waiting on this to go live, so hopefully we can distribute today or tomorrow.


derrickstaten:
okay this is much cleaner

Kirkkt:
:+1:

PR #153, opened 2015-08-26
reapply PR #143, fix referring_link doesn't persist

Kirkkt:
plus 2 jshint rules that we are already following (which means it doesn't break lint)


dmitrig01:
:+1:

Kirkkt:
@dmitrig01 

Once this is in we are ready for a new release with two major bug fixes :)

PR #155, opened 2015-08-26
indent example html properly

Kirkkt:
@dmitrig01

This file is basically a potato soup of tabs and spaces and bad indentations. we can't lint it because it's html file by nature.


Kirkkt:
ping @dmitrig01 

FYI this is an NOP PR.

dmitrig01:
:+1:

PR #156, opened 2015-08-26
use square bracket because it messes up the minification

Kirkkt:
@dmitrig01 @derrickstaten 

What is happening here is we are using dot notation and it passes all tests. However, the minification messes it up by trying to rename it to a one-letter attribute to save space. End result? the cdn version (which has been minified) isn't working because of this.


derrickstaten:
:+1:

PR #157, opened 2015-08-27
Make sure we always test against the minified version locally

Kirkkt:
@dmitrig01


dmitrig01:
:+1: but this won't run tests, only example.html

PR #158, opened 2015-08-27
re-make integration-test.html to test against min.js

Kirkkt:
@dmitrig01 sorry I forgot to re-make this


Kirkkt:
oops we can't just do this because test/4_integration.js speaks the unminified language :(

PR #160, opened 2015-08-27
update grunt to include linux, more variety of testing env and follow the saucelabs configuration tool strictly

Kirkkt:
also more variety of testing env and follow the saucelabs configuration tool strictly: https://docs.saucelabs.com/reference/platforms-configurator/#/

@dmitrig01


dmitrig01:
:+1:

PR #161, opened 2015-08-27
JS strict mode (with two minor fixes) + disable grunt test locally

Kirkkt:
@dmitrig01 

Let's settle this once and for all!


Kirkkt:
The package.json effect of this PR is:

lint.sh will run locally
grunt test won't run locally

src/3_branch.js
```diff
@@ -106,7 +107,7 @@ Branch = function() {
 		}
 		return default_branch;
 	}
-	this._queue = Queue();
+	this._queue = new Queue();
```
dmitrig01: this will break?

Kirkkt: nice catch!

Kirkkt: It's caused by bad naming, this `Queue` is not a constructor, but when named upper-camelly, jshint thinks it is.

Renamed to task_queue (queue doesn't work because it introduces another global-local variable name conflict)

dmitrig01:
:+1:

PR #165, opened 2015-08-30
[SIMPLE] remake after adding a comma in comments

Kirkkt:
oops I forgot to run `make` :P


dmitrig01:
i don't think we need to do PRs for remakes, but better safe than sorry I suppose :-)

PR #168, opened 2015-08-31
make test/0_queue.js order independent

Kirkkt:
@dmitrig01 

This test has an implicit assumption of orders, which is that the first test case being run before the second. This is not a good assumption when we run unit tests (for example: running the second test by itself will fail).


dmitrig01:
great catch :+1:

PR #169, opened 2015-08-31
remove the useless creation_type from getCode()

Kirkkt:
@dmitrig01 

`creation_type` is nowhere to be found in the docs, and it's nowhere to be found in our API. However, web sdk refuses to issue an API call unless this parameter is specified. 

What might have happened, I guess, is that 1, at one point there was a parameter called creation_type; later the backend changed but the web sdk forgot to drop this; or 2, it's merely a typo.

As a result, the server simply ignore this parameter like some nonsense in silence. The problem we are seeing here is the result of the web sdk insisting on having this parameter.

I think it's safe to just remove it. Will let dmitri make the call :)

cc @rubinsingh


Kirkkt:
friendly ping
@dmitrig01

dmitrig01:
actually - can you look at creation_source in the backend? see if promo codes use that? if so, just rename it to that?

Kirkkt:
Brilliant!

Kirkkt:
renamed to `creation_source` 

@dmitrig01

dmitrig01:
:+1:

PR #174, opened 2015-09-02
[SIMPLE][NOP] use underscore for markdown horizontal separators

Kirkkt:
@dmitrig01 

We use `* * *`, `---` and `___` all over the place. This PR consolidate them under one ring.


dmitrig01:
:+1:

Kirkkt:
Oh forgot to mention that there is no visible diff.

PR #180, opened 2015-09-11
[SIMPLE] Richer error message on assert.equal()

Kirkkt:
Before this PR, when assert(a, b, "message") fails we only see "message"

After this PR, we also see the_value_of_a and the_value_of_b.

Basically a nop. Helpful for debugging

@dmitrig01


dmitrig01:
:+1:

PR #181, opened 2015-09-11
Use string attribute for browser_fingerprint_id

Kirkkt:
It's safer than dot notation against minification

@dmitrig01


dmitrig01:
oo yeah good catch :+1:. possible to add tests for this?

Kirkkt:
Great point! Allow me to do this later. Right now I'd like to do the absolute minimal to get master back to green so that we can stand on a solid ground before adding additional logic (including tests).

Let me know if you think otherwise :)

chanian:
:+1:

PR #182, opened 2015-09-11
stop race conditions in test/0_storage

Kirkkt:
The issue here is when we test "should get stored item with key", we assume storage.set() has been called, and this order is not always guaranteed.

This PR ensures every individual test runs in a silo

@dmitrig01


dmitrig01:
:+1:

chanian:
:+1:

PR #183, opened 2015-09-11
Make api and branch unit tests green (except for firefox browsers)

Kirkkt:
A few flakiness-prone issues:

**violation of encapsulation**:
`clock = sinon.useFakeTimers();` in test/0_queue.js is used across test cases.
`storage` and `requests` are used across test cases.

**double restoration**:
sinon stub restore() may cause failures if sinon tries to restore a stub that has already been restored

**stubbing undefined**:
sinon stub() may cause failures if the function to be stubbed is undefined

**stop stubbing other people's stuff**:
We shouldn't stub `goog.json.parse()`, use the proper way to check for empty object instead!

So uh.... what's up with firefox?
Firefox browsers seem to have a whole different deal. I will investigate in the next PR.

@dmitrig01 @chanian


test/1_api.js
```diff
@@ -119,6 +120,7 @@ describe('Server', function() {
 		storage.clear();
 		xhr = sinon.useFakeXMLHttpRequest();
 		clock = sinon.useFakeTimers();
+		server.createScript = function() {};
```
dmitrig01: is this one ever undefined?

test/1_api.js
```diff
@@ -130,20 +132,21 @@ describe('Server', function() {
 	});
 
 	afterEach(function() {
-		xhr.restore();
-		clock.restore();
-		server.createScript.restore();
+		if (typeof xhr.restore === 'function') {
```
dmitrig01: when would these guys happen?

test/1_api.js
Kirkkt: It's safer. Since we are stubbing it anyway, I'd prefer to nullify it here so that we don't have to worry about it.

test/1_api.js
Kirkkt: https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=Attempted+to+wrap+which+is+already+wrapped

To be honest I only know that this causes flakiness but I don't know why. I like this check for safety reasons.

dmitrig01:
:+1:

test/1_api.js
Kirkkt: To answer your question directly, I don't remember seeing it happened in the past, but I've seen instances of, when we try to stub `obj.func`, some browsers on saucelabs (I'm unable to reproduce it locally) will yell "attempted to stub undefined" on `obj.func`

dmitrig01: weird. ok:-)

Kirkkt: the way I see it, flakiness -> indeterminism -> weirdness, so all these flakiness fixes are pretty weird

I can't wait to fix integration tests. It will be a freakshow for show.

sign...

:panda_face:

PR #184, opened 2015-09-12
Re-enable safari + safeguard stub and restore for integration tests

Kirkkt:
@dmitrig01 

Safari sometimes freaks out on this. Safety, safety :)


test/4_integration.js
```diff
@@ -73,14 +75,18 @@ describe('Integration tests', function() {
 	});
 
 	after(function() {
-		if (window.WEB_BUILD) {
+		if (window.WEB_BUILD && typeof branch._server.createScript.restore === 'funciton') {
```
dmitrig01: typo :-) curious that this passes tests though!

Kirkkt: Nice catch! 
image

The reason it's passing the tests is that it's in after() so the function doesn't get restored when all the test cases are processed, but we no longer have more test cases by then. It's still a good idea to restore, though.

The reason it's passing lint is ... uh ... I turned it off because I committed very many times today. Turning it back on since we are close to finish!

dmitrig01:
:+1:

PR #187, opened 2015-09-13
Put back previously failed test case in 0_utils and shorten lines to below 100c

Kirkkt:
I initially only wanted to put back the test case `should make missing params null`.

Unfortunately the previous PR made some lines too long so I fixed them here as well.

Pardon the mess! 
@dmitrig01


dmitrig01:
:+1:

PR #189, opened 2015-09-14
[NOP] test/3_branch.js readability improvement

Kirkkt:
1, assure assert.equal(actual, expected, message) order
2, better line breaks


Kirkkt:
Oops, forgot to ping you :) @dmitrig01

dmitrig01:
:+1:

PR #191, opened 2015-09-14
[SIMPLE] this console.log doesn't seem to do anything

Kirkkt:
also inlined a var since... why bother?

@dmitrig01


dmitrig01:
I would prefer to keep constants separate from code - but perhap we should change it to COOKIE_MS or something?

dmitrig01:
that way we don't have any 'magic numbers', but instead they are labelled

Kirkkt:
cool idea!

Kirkkt:
Done :)

@dmitrig01

PR #193, opened 2015-09-14
[SIMPLE] readability: reduce level of indents for branch.init()

Kirkkt:
I'm about to make some non-trivial changes but found this function a pain to read/edit. Essentially an NOP.

@dmitrig01


Kirkkt:
Might as well do this across the board and set a lint rule for it :)

Kirkkt:
friendly ping
@dmitrig01

src/3_branch.js
```diff
@@ -446,82 +446,84 @@ Branch.prototype['init'] = wrap(
...
+			return;
 		}
-		else {
-			if (CORDOVA_BUILD || TITANIUM_BUILD) {
 
-				var apiCordovaTitanium = function(data) {
-					if (!freshInstall) {
-						data['identity_id'] = sessionData['identity_id'];
-						data['device_fingerprint_id'] = sessionData['device_fingerprint_id'];
+		if (WEB_BUILD) {
```
dmitrig01: i'm pretty sure we still need an else, no? we can definitely turn it into an elseif though

Kirkkt: it returns in the bottom

dmitrig01: ah ok cool

dmitrig01:
:+1:

dmitrig01:
actually hang on

dmitrig01:
you sure if resources.queryPart is undefined that loop won't throw an error (in all browsers)?

Kirkkt:
@dmitrig01 Great point! I'm not 100% sure so it's better safe than sorry.

Thank you!

dmitrig01:
:+1:

if you can check this, we do this in a lot of places :-)

Kirkkt:
Done :)

dmitrig01:
nice :+1:

PR #194, opened 2015-09-16
[SIMPLE] readable onpage.js

Kirkkt:
1, stop splitting a string since array is what we want naturally
2, sort
3, indent

@dmitrig01


dmitrig01:
to be honest, the latter is more readable to me. I think unless we have a standard for this, it's not worth wasting time making these kinds of changes.

btw, if you'll notice, the minified version doesn't actually change (except for the order). if you feel strongly about making it an array, we can do that, but again, I don't feel strongly and as such would be generally against wasting time here

Kirkkt:
Cool. I understand. Since I've already made this PR, shall we proceed with this and I will refrain from making similar PRs in the future?

I like the sorting part because in the future if we have many functions it'll be much less cognitive load to read through them.

Thanks for your comment!

dmitrig01:
sounds good.

If you want to establish a stricter coding standard then I would be happy accepting more PRs that change code style :-)

PR #195, opened 2015-09-16
introduce dynamically inject js function in 0_utils

Kirkkt:
@dmitrig01 

This is the first step towards deepview mode.

I personally prefer sending out multiple small PRs over aggregating a big one.

This new function is necessary for us to embed the sendToApp and sendToStore functions we get from the backend, and it can be extracted into a standalone PR.

Let me know if you think otherwise on any of the above :)

### DIFF
--- a/src/0_utils.js
+++ b/src/0_utils.js
@@ -263,3 +263,16 @@ utils.base64encode = function(input) {
 	}
 	return output;
 };
+
+/**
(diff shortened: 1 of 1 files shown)


dmitrig01:
it'd be great if you could make one big pr (as long as it's <1000 lines), that way i could review a function like this in the context of how it's called

dmitrig01:
thanks!

PR #198, opened 2015-09-22
send browser finegrprint on every request

dmitrig01:
@Kirkkt


Kirkkt:
test/test.html seems broken, can you double-check?

dmitrig01:
it thorws some errors for me, but they don't seem to be related - i see errors with the history pushstate - but i get those on master too

Kirkkt:
all tests passed on the latest master branch for me

for starters, can you doublecheck whether you need to update this test https://github.com/BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK/blob/send-browser-fingerprint-id/test/6_branch.js#L400 ?

dmitrig01:
weird - those are not failing on me, but they definitely would need to be changed. hmm, something is definitely fishy here

dmitrig01:
fixed

Kirkkt:
:+1:
sick

PR #199, opened 2015-09-23
[FIX] Allow web sdk to be called after prior requests error out

Kirkkt:
The issue here is: after a request results in an error, e.g. no network connection, web SDK refuses to take any additional incoming method calls.

The cause is that we didn't set req.onerror()

@dmitrig01


src/3_api.js
```diff
@@ -255,6 +252,9 @@ Server.prototype.XHRRequest = function(url, data, method, storage, callback) {
 		};
 	}
 	else {
+		req.onerror = function(e) {
```
dmitrig01: what's the cross-platform implication of this?

dmitrig01:
if there's an error init'ing, will it block all further calls? or will a subsequent call make the init retry?

can we get some tests around this?

dmitrig01:
also there are still some console.logs in here :-)

Kirkkt:
@dmitrig01
 
luckily for init we keep retrying (on apr/_r), and when the user reconnect the init can finish without an error

as for cross-platform, the unit tests all passed

And I removed the silly console logs

dmitrig01:
this is great

can we write another test that ensures the retry, i.e. if you call branch.init() and it fails, then you call branch.link(), the latter call will re-try the init call and then if that succeeds, call link?

dmitrig01:
After thinking it over, this PR on it's own _should_ be fine - but let's go over it with @aaustin

dmitrig01:
:+1:

PR #200, opened 2015-09-24
[NOP]sort two json

Kirkkt:


Kirkkt:
we also need to update Makefile to parse package.json differently... not worth it

PR #201, opened 2015-09-24
Sanity check the build flag

Kirkkt:
Sanity check the build flag


Makefile
```diff
@@ -25,7 +25,7 @@ ONPAGE_TEST=$(subst ",\",$(shell perl -pe 'BEGIN{$$sub="../dist/build.js"};s\#SC
 
 .PHONY: clean
 
-all: dist/build.min.js dist/build.js README.md TITANIUM_GUIDE.md WEB_GUIDE.md example.html test/branch-deps.js dist/build.min.js.gz dist/build.min.js.gz test/integration-test.html
+all: dist/build.min.js dist/build.js README.md TITANIUM_GUIDE.md WEB_GUIDE.md example.html test/branch-deps.js test/integration-test.html
```
Kirkkt: unrelated change: let's not bother with .gz unless we are making a release

package.json
```diff
@@ -27,8 +27,6 @@
 		"grunt-saucelabs": "~8.6.1",
 		"jsdoc": "3.2.2",
 		"jsdox": "0.4.9",
-		"karma": "^0.12.37",
```
Kirkkt: we are not using karma

Kirkkt:
@dmitrig01 

A handful of miscellaneous things. Please see inline comments for details.

src/0_config.js
```diff
@@ -15,3 +15,10 @@ var CORDOVA_BUILD = false; // jshint unused:false
 
 /** @define {boolean} */
 var TITANIUM_BUILD = false; // jshint unused:false
+
+var isApp = (document.URL.indexOf('http://') === -1) && (document.URL.indexOf('https://') === -1);
```
Kirkkt: Sanity check the build flag

dmitrig01: looks great. 

1) i might do an alert, although
2) i would rather check window.cordova instead - this would prevent anyone from running it outside the context of http

dmitrig01: we should also add these propeties to the exerns file

dmitrig01: *externs

Kirkkt:
Not sure if this is what you had in mind, PTAL @dmitrig01

dmitrig01:
:+1:

PR #206, opened 2015-09-30
Fix syntax for sendSMS example

steaks:


derrickstaten:
the readme is derived, so we'll need to fix in the src file. incorporated into https://github.com/BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK/pull/207

thank you so much for bringing this to our attention!

PR #207, opened 2015-09-30
fix example - don't stringify err

derrickstaten:
@dmitrig01 this does not include the request header change needed to trigger the `js_response` code path on link-service

### DIFF
--- a/src/6_branch.js
+++ b/src/6_branch.js
@@ -922,7 +922,7 @@ Branch.prototype['link'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
  * ##### Example
  * ```js
  * branch.sendSMS(
- *     phone: '9999999999',
+ *     '9999999999',
  *     {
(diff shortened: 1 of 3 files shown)


dmitrig01:
you need to edit example.template.html and then make

derrickstaten:
@dmitrig01 updated

dmitrig01:
:+1:

src/6_branch.js
```diff
@@ -922,7 +922,7 @@ Branch.prototype['link'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
  * ##### Example
  * ```js
  * branch.sendSMS(
- *     phone: '9999999999',
+ *     '9999999999',
```
Kirkkt: good! please run `make` after changing this, which will update docs/web/3_branch_web.md

Kirkkt:
:+1: except the above knit

PR #211, opened 2015-10-19
[initial review] deepview init

Kirkkt:
what is included:
when deepview is init-ed, fetch the embedded js and run it
use the deepview CTA links for banner

what is not included:
tests

@dmitrig01 @derrickstaten


src/6_branch.js
```diff
@@ -541,6 +541,39 @@ Branch.prototype['init'] = wrap(
 );
 
 /**
+ * @function Branch.deepview
+ * @param {function(?Error, utils.sessionData=)=} callback - _optional_ - callback to read the session data.
```
dmitrig01: we wnat the callback to accept a function/url for the CTA, right? or do we want a separate/always global function to handle a CTA click?

src/6_branch.js
```diff
@@ -541,6 +541,39 @@ Branch.prototype['init'] = wrap(
...
+			referringLink.lastIndexOf('/') + 1, referringLink.length
+		);
+	}
+
+	this._api(resources.deepview, cleanedData, function(err, data) {
+		if (err) {
+			done(err, data);
+		}
+
+		self._server.createScript(data[0], true /* internal */);
```
dmitrig01: ah - we can do this in an even simpler way: just treat the API request as a jsonp request, that way it gets put in a JS embed. then we just have to be sure to actually call the callback in the js that's returned from the server. does that make sense?

Kirkkt: done :)

src/6_branch.js
Kirkkt: I haven't added docs yet, is it okay to postpone it for now?

Kirkkt:
@dmitrig01 PTAL

src/6_branch.js
```diff
@@ -541,6 +541,56 @@ Branch.prototype['init'] = wrap(
...
+
+		this._api(resources.deepview, cleanedData, function(err, data) {
+			if (err) {
+				return done(err);
+			}
+
+			if (typeof data === 'function') {
+				Branch.prototype['deepviewCta'] = data;
+			}
+			else {
```
dmitrig01: when is this the case?

Kirkkt: calling the callback from the handlebars js -> if
the normal callback -> else

src/6_branch.js
```diff
@@ -541,6 +541,56 @@ Branch.prototype['init'] = wrap(
...
+		}
+
+		this._api(resources.deepview, cleanedData, function(err, data) {
+			if (err) {
+				return done(err);
+			}
+
+			if (typeof data === 'function') {
+				Branch.prototype['deepviewCta'] = data;
+			}
```
dmitrig01: we need to still call done()

src/6_branch.js
```diff
@@ -541,6 +541,56 @@ Branch.prototype['init'] = wrap(
...
+			);
+		}
+
+		this._api(resources.deepview, cleanedData, function(err, data) {
+			if (err) {
+				return done(err);
+			}
+
+			if (typeof data === 'function') {
+				Branch.prototype['deepviewCta'] = data;
```
dmitrig01: can we make deepviewCta a top-level branch function, and then assign this to like this._deepviewCta, and then if the top-level deepviewCta doesn't have a "this._deepviewCta", we can throw some error like "cannot trigger deepview cta if no deepview"

src/6_branch.js
```diff
@@ -1583,7 +1633,8 @@ if (WEB_BUILD) {
 			desktopSticky: typeof options['desktopSticky'] === 'undefined' ?
 				true :
 				options['desktopSticky'],
-			make_new_link: !!options['make_new_link']
+			make_new_link: !!options['make_new_link'],
+			open_app: !!options['open_app']
```
dmitrig01: indentation wrong

dmitrig01: never mind

src/6_branch.js
dmitrig01: no, let's add the docmentation. should be simple

src/6_branch.js
Kirkkt: no we don't. because this will cause two branch.deepview callbacks. 

for example, if the user does `branch.deepview(my_data, my_callback);` `my_callback` will be called twice.

src/6_branch.js
dmitrig01: when does the normal callback happen?

Kirkkt: when the api calls `res.status(200).jsonp(...)` so the calling of the callback from the handlebars js happens first, and then the normal callback

dmitrig01: api should not call jsonp. it should call res.status(200).send, and _always_ return jsonp

src/6_branch.js
Kirkkt: Done

src/6_branch.js
Kirkkt: done

src/6_branch.js
```diff
@@ -1027,6 +1027,170 @@ Branch.prototype['sendSMS'] = wrap(
...
+ * deepview finishes, and thus the CTA is to visit the platform-appropriate app stores; if on the
+ * other hand, the deepview is configured with the option *`open_app`* being false, the CTA is to
+ * try to open app, and to visit the platform-appropriate app stores if the open-app attempt failes.
+ *
+ * If *branch.deepview* has not been called, an error will arise with a reminder to call
+ * *branch.deepview* first.
+ * ___
+ */
+/*** +TOC_ITEM #deepview-callback &.deepviewCta()& ^ALL ***/
+Branch.prototype['deepviewCta'] = wrap(callback_params.CALLBACK_ERR, function(done) {
```
dmitrig01: this one shouldn't need to take a callback i don't think

Kirkkt: done. good point

src/6_branch.js
Kirkkt: done

Kirkkt:
unit tests added, will add integration tests in a separate PR

@dmitrig01

src/6_branch.js
```diff
@@ -1027,6 +1016,187 @@ Branch.prototype['sendSMS'] = wrap(
...
+ *
+ * These reward definitions are created on the dashboard, under the 'Reward Rules' section in the 'Referrals' tab on the dashboard.
+ *
+ * Warning: For a referral program, you should not use unique awards for custom events and redeem pre-identify call. This can allow users to cheat the system.
+ *
+ * ## Retrieve referrals list
+ *
+ */
+/*** +TOC_ITEM #deepviewctaperform &.deepviewCta()& ^ALL ***/
+Branch.prototype['deepviewCta'] = wrap(callback_params.CALLBACK_ERR, function(done) {
```
dmitrig01: again, don't need a callback here

dmitrig01: i think we don't need to use wrap() at all, but i'm not sure

src/6_branch.js
```diff
@@ -1027,6 +1016,187 @@ Branch.prototype['sendSMS'] = wrap(
...
+ * deepview finishes, and thus the CTA is to visit the platform-appropriate app stores; if on the
+ * other hand, the deepview is configured with the option *`open_app`* being false, the CTA is to
+ * try to open app, and to visit the platform-appropriate app stores if the open-app attempt failes.
+ *
+ * If *branch.deepview* has not been called, an error will arise with a reminder to call
+ * *branch.deepview* first.
+ *
+ * ##### Usage
+ * ```js
+ * branch.deepviewCta();
```
dmitrig01: we may want to use an example with a link, since the idea is to put this behind a link. for example,

```
// If you're using jQuery, then:
$('a.deepview-cta').click(function() { branch.deepviewCta(); return false });
```

Now that I think of it, we may want to allow users to bind directly to a link like this:
```
$('a.deepview-cta').click(branch.deepviewCta);
```

or 

```
document.getElementById('my-elem').onClick = branch.deepviewCta;
```

in which case we'll either want to have deepviewCta take the click event as an argument, and call `.preventDefault()`, or we'll want to return false (o maybe both).

Though we shouldn't require it to be called like that - we should allow people to call it whenever they want. So we need to verify that preventDefault is callable and such.

src/6_branch.js
```diff
@@ -1027,6 +1016,187 @@ Branch.prototype['sendSMS'] = wrap(
...
+
+		this._api(resources.deepview, cleanedData, function(err, data) {
+			if (err) {
+				return done(err);
+			}
+
+			if (typeof data === 'function') {
+				self._deepviewCta = data;
+			}
+			else {
```
dmitrig01: as with before, we don't need this second branch

src/6_branch.js
```diff
@@ -1027,6 +1016,187 @@ Branch.prototype['sendSMS'] = wrap(
...
+	function(done, data, options) {
+		var self = this;
+
+		if (!options) {
+			options = { };
+		}
+		var cleanedData = utils.cleanLinkData(data);
+
+		if (cleanedData['tags']) {
+			cleanedData['tags'] = goog.json.serialize(cleanedData['tags']);
```
dmitrig01: i don't think there's a particular reason to do this, is there?

src/6_branch.js
```diff
@@ -1027,6 +1016,187 @@ Branch.prototype['sendSMS'] = wrap(
...
+		}
+
+		if (cleanedData['data']) {
+			cleanedData['metadata'] = cleanedData['data'];
+			delete cleanedData['data'];
```
dmitrig01: any reason to do this and not call it 'data', since we use 'data' everywhere else?

Kirkkt: app/index.js does this parsing thing for req.query.data for all incoming GET requests, which has some base64blahblah logic that I don't want to further complicate just for deepview's sake, and since we only expect users to call the deepview api endpoint directly, I figure I'd just pass in metadata, instead of data, to bypass the base64blahblah parsing (if we pass in data like this, it's gonna yell "unable to parse data").

In short, it was a lazy hack so that I can piggyback on utils.cleanLinkData() here without having to bother with the more complicated stringification/parsing mechanisms.

src/6_branch.js
Kirkkt: you are correct. Removed

src/6_branch.js
Kirkkt: you are totally right!
goodness me, how can you "run" it in your head like that...

src/6_branch.js
Kirkkt: done. let me know if I made this function unnecessarily complicated

src/6_branch.js
Kirkkt: done, sorry I forgot

Kirkkt:
@dmitrig01 PTAL

src/6_branch.js
```diff
@@ -1027,6 +1016,210 @@ Branch.prototype['sendSMS'] = wrap(
...
+	function(done, data, options) {
+		var self = this;
+
+		if (!options) {
+			options = { };
+		}
+		var cleanedData = utils.cleanLinkData(data);
+
+		if (cleanedData['data']) {
+			cleanedData['metadata'] = cleanedData['data'];
```
dmitrig01: somehow this lost my comment here - is there a reason to call it metadata instead of data?

src/6_branch.js
dmitrig01: ah, i see. ahh, yknow what we can do - if you set it to be a "POST request" over jsonp, it'll put *everything* inside data, and the api should parse that out automatically

Kirkkt: You are right, it used to work flawlessly under POST until I changed it to GET. but do you feel strong about this? I'd advocate just using metadata since web sdk is the only client as we can see for now; if future demands arrive, we can change it later.

Kirkkt: just to keep things simple. This is my laziness talks. Let me know if you think otherwise.

dmitrig01: cna you try setting the method to post, but keeping the jsonp?

dmitrig01: it may just work, and allow you to call it data instead of metadata. that way even though it's just the web sdk client, it'll allow us to be consistent

Kirkkt: done :D

Kirkkt:
@dmitrig01 PTAL

src/3_api.js
```diff
@@ -163,11 +165,18 @@ Server.prototype.getUrl = function(resource, data) {
  * This function is standalone for easy mocking.
  * @param {string} src
  */
-Server.prototype.createScript = function(src) {
+Server.prototype.createScript = function(src, internal) {
```
dmitrig01: i don't think we use it with internal: true, do we?

Kirkkt: good catch, this PR no longer does this. remove in just a bit

Kirkkt:
@jsaleigh

src/6_branch.js
```diff
@@ -1027,6 +1016,201 @@ Branch.prototype['sendSMS'] = wrap(
...
+ */
+/*** +TOC_ITEM #deepviewctaevent-silent &.deepviewCta()& ^ALL ***/
+Branch.prototype['deepviewCta'] = function(event, silent) {
+	if (typeof this._deepviewCta === 'undefined') {
+		if (silent) {
+			return true;
+		}
+		else {
+			throw new Error(
+				'Cannot call deepview CTA, did you forget to call branch.deepview()?'
```
dmitrig01: let's phrase this a little more nicely as "Cannot call Deepview CTA, please call branch.deepview() first"

src/6_branch.js
```diff
@@ -1027,6 +1016,201 @@ Branch.prototype['sendSMS'] = wrap(
...
+ * These reward definitions are created on the dashboard, under the 'Reward Rules' section in the 'Referrals' tab on the dashboard.
+ *
+ * Warning: For a referral program, you should not use unique awards for custom events and redeem pre-identify call. This can allow users to cheat the system.
+ *
+ * ## Retrieve referrals list
+ *
+ */
+/*** +TOC_ITEM #deepviewctaevent-silent &.deepviewCta()& ^ALL ***/
+Branch.prototype['deepviewCta'] = function(event, silent) {
+	if (typeof this._deepviewCta === 'undefined') {
```
dmitrig01: we may want to look at putting this in the queue - see branch.data - that way if the deepview call is in progress, deepviewCta will wait until that call is done

Kirkkt: good idea! done.

src/6_branch.js
Kirkkt: done.

Kirkkt:
@dmitrig01 Thanks for the insights :) PTAL

WEB_GUIDE.md
```diff
@@ -468,6 +470,148 @@ callback("Error message");
...
+
+// Or in HTML you can:
+<a href='...' onclick='branch.deepviewCta(null, true); return false'>
+
+// If you wish to dedicate a CTA link only to branch deepviewCta, you can:
+branch.deepview(data, option, function(err) {
+    if (err) {
+        throw err;
+    }
+    ${'a.deepview-cta').click = branch.deepviewCta(null, false);
```
dmitrig01: if you're using jquery you'd need to do `.click(branch.deepviewCta);`

WEB_GUIDE.md
```diff
@@ -468,6 +470,148 @@ callback("Error message");
...
+##### Usage
+```js
+// If you're using jQuery, then:
+$('a.deepview-cta').click(branch.deepviewCta);
+
+// Or you can:
+document.getElementById('my-elem').onClick = branch.deepviewCta;
+
+// Or in HTML you can:
+<a href='...' onclick='branch.deepviewCta(null, true); return false'>
```
dmitrig01: you don't need to return false since branch.deepviewCta already returns false. i think it's fine to remove the second param also

src/6_branch.js
```diff
@@ -1027,6 +1016,205 @@ Branch.prototype['sendSMS'] = wrap(
...
+ *
+ * These reward definitions are created on the dashboard, under the 'Reward Rules' section in the 'Referrals' tab on the dashboard.
+ *
+ * Warning: For a referral program, you should not use unique awards for custom events and redeem pre-identify call. This can allow users to cheat the system.
+ *
+ * ## Retrieve referrals list
+ *
+ */
+/*** +TOC_ITEM #deepviewctaevent-silent &.deepviewCta()& ^ALL ***/
+Branch.prototype['deepviewCta'] = wrap(callback_params.NO_CALLBACK, function(done, event, silent) {
```
dmitrig01: i believe i already left this comment, but since we are not expecting a callback, i don't think we need to use wrap at all

src/6_branch.js
```diff
@@ -1027,6 +1016,205 @@ Branch.prototype['sendSMS'] = wrap(
...
+	if (typeof this._deepviewCta === 'undefined') {
+		if (silent) {
```
dmitrig01: what's the thinking behind silent? when would you want to set it?

src/6_branch.js
```diff
@@ -1027,6 +1016,205 @@ Branch.prototype['sendSMS'] = wrap(
...
+			done();
```
dmitrig01: i think we'll have to call done every time, otherwise the queue will get jamemd

dmitrig01: *jammed

src/6_branch.js
dmitrig01: actually never mind - this will put it in the queue, which means that if you run branch.deepview, this will necessarily wait until after it's done - which is exactly what we want

src/6_branch.js
Kirkkt: if you are a lazy developer who doesn't want to set the action handler to deepviewCta inside deepview callback, instead you prefer to do `onclick="branch.deepviewCta()"` inside HTML directly, and you don't want to handle the "call deepview first before you call deepviewCta" error, then you sent `silent` to true and if the user clicks the CTA before deepview() call is finished, it will fail silently.

Or do you think we should always threw error when deepview() call is not finished?

dmitrig01: i think we should always throw an error

dmitrig01: but besides, since we're putting the deepviewCta call on the queue, as long as you call deepview before deepviewCta, even if the deepview call doesn't finish, deepviewCta will wait for it to finish

WEB_GUIDE.md
Kirkkt: done. thanks!

WEB_GUIDE.md
Kirkkt: done

src/6_branch.js
Kirkkt: done, makes sense

src/6_branch.js
Kirkkt: done

Kirkkt:
also got rid of temporary hacks

dist/build.js
```diff
@@ -721,7 +721,7 @@ goog.json.Serializer.prototype.serializeObject_ = function(a, b) {
   b.push("}");
 };
 // Input 2
-var config = {link_service_endpoint:"https://bnc.lt", api_endpoint:"https://api.branch.io", version:"1.7.1"}, WEB_BUILD = !0, CORDOVA_BUILD = !1, TITANIUM_BUILD = !1, IS_CORDOVA_APP = !!window.cordova;
+var config = {link_service_endpoint:"http://localhost:5002", api_endpoint:"http://localhost:5001", version:"1.7.1"}, WEB_BUILD = !0, CORDOVA_BUILD = !1, TITANIUM_BUILD = !1, IS_CORDOVA_APP = !!window.cordova;
```
jsaleigh: It looks like a debug config snuck through with localhost

dist/build.js
```diff
@@ -1109,7 +1110,7 @@ Server.prototype.createScript = function(a) {
 };
 var jsonp_callback_index = 0;
 Server.prototype.jsonpRequest = function(a, b, c, d) {
-  var e = "branch_callback__" + this._jsonp_callback_index++, f = 0 <= a.indexOf("api.branch.io") ? "&data=" : "&post_data=";
+  var e = "branch_callback__" + this._jsonp_callback_index++, f = 0 <= a.indexOf("api.branch.io") || 0 <= a.indexOf("localhost:5001") ? "&data=" : "&post_data=";
```
jsaleigh: Possible debug config, or is this what is used for direct links on the server?

dist/build.js
Kirkkt: I forgot to run `make`, done.

dist/build.js
Kirkkt: ditto, done.

Kirkkt:
@dmitrig01 PTAL

Kirkkt:
@dmitrig01 thoughts on this?

WEB_GUIDE.md
```diff
@@ -466,17 +468,153 @@ callback("Error message");
...
+&#x2F;&#x2F; If you wish to dedicate a CTA link only to branch deepviewCta, you can:
+branch.deepview(data, option, function(err) {
+    if (err) {
+        throw err;
+    }
+    ${&#39;a.deepview-cta&#39;).click(branch.deepviewCta);
+});
+
+&#x2F;&#x2F; You can call this function any time after branch.deepview() is finished by simply:
+branch.deepviewCta();
```
aaustin: If this fails for a network call if we're down or something. It's a pretty crappy developer/user experience. Can the fail case click an /a/ link created from the inputs of deepview(). Here's how to construct an /a/ link: https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/blob/master/Branch-SDK/Branch-SDK/Branch.m#L1072

aaustin: I say this because this is what we do if the request fails with Pinterest. Otherwise, their website doesn't work if the Branch API is down.

Kirkkt: This proposal only works when branch api is down and branck link-service is up, right? Do we want to prepare for this somewhat improbable case?

dmitrig01: this has happened several times before

aaustin: If I'm a dev, I would simply make all my buttons call deepviewCta. If there is an init blip that causes the web sdk to fail to load, at least the the bnc.lt link would still be formatted. I wouldn't know if the user action actually failed or not.

dmitrig01: I think it's valid to require a call to deepviewCta after a call to deepview - you should be calling deepview on page load - but I think we can say neither in it nor deepview have to _succeed_ before calling deepviewCta

aaustin: Agreed. Or we need a failure callback on deepviewCta where the dev can handle that network blip case.

Kirkkt: sounds good, I will add an a/ link fallback

aaustin:
We'll need to update the init snippet as well for people to be able to use it.

dmitrig01:
that's done here

Reference.md
```diff
@@ -391,6 +393,140 @@ callback("Error message");
...
+If *branch.deepview* has not been called, an error will arise with a reminder to call
+*branch.deepview* first.
+
+##### Usage
+```js
+$('a.deepview-cta').click(branch.deepviewCta); // If you are using jQuery
+
+document.getElementById('my-elem').onClick = branch.deepviewCta; // Or generally
+
+<a href='...' onclick='branch.deepviewCta()'> // In HTML
```
aaustin: We probs don't want them to have an `href=` if they are overriding onclick with Branch stuff.

Kirkkt:
@aaustin please ignore `WEB_GUILD.md`, it's no longer correctly generated by Makefile in the midst of the documentation overhaul effort, `Reference.md` is the new cool kid in school

Reference.md
Kirkkt: agree, thanks!

Kirkkt:
@aaustin @dmitrig01 

PTAL, I will add unit tests if it looks good.

Kirkkt:
more unit tests needed

Kirkkt:
more unit tests added, @aaustin @dmitrig01 PTAL

dmitrig01:
@jsaleigh would love a review here. if you need a walkthrough of the functionality i'm happy to do so

src/2_resources.js
```diff
@@ -334,6 +335,23 @@ resources.link = {
 	})
 };
 
+resources.deepview = {
+	destination: config.api_endpoint,
+	endpoint: "/v1/deepview",
+	jsonp: true,
+	method: utils.httpMethod.POST,
+	params: defaults({
```
jsaleigh: I don't see this verb for this endpoint, unless I'm missing it?

dmitrig01: method: utils.httpMethod.POST

WEB_GUIDE.md
```diff
@@ -466,17 +468,153 @@ callback("Error message");
...
+
+
+### Branch.deepviewCta
+
+Perform the branch deepview CTA (call to action) on mobile. Namely, depends on how
+*branch.deepview* is set up, the mobile users are redirected accordingly. If the deepview is
+configured with the option *`open_app`* being true, an immediate attempt is made as soon as
+deepview finishes, and thus the CTA is to visit the platform-appropriate app stores; if on the
+other hand, the deepview is configured with the option *`open_app`* being false, the CTA is to
+try to open app, and to visit the platform-appropriate app stores if the open-app attempt failes.
```
jsaleigh: typo 'failes'

src/2_resources.js
```diff
@@ -334,6 +335,23 @@ resources.link = {
...
+		"campaign": validator(false, validationTypes.STRING),
+		"channel": validator(false, validationTypes.STRING),
+		"data": validator(true, validationTypes.STRING),
```
jsaleigh: Should this be validationTypes.OBJECT?

dmitrig01: yes

src/6_branch.js
```diff
@@ -983,23 +983,7 @@ Branch.prototype['link'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done
...
- *
- * These reward definitions are created on the dashboard, under the 'Reward Rules' section in the
- * 'Referrals' tab on the dashboard.
- *
- * Warning: For a referral program, you should not use unique awards for custom events and redeem
- * pre-identify call. This can allow users to cheat the system.
- *
- * ## Retrieve referrals list
- *
+ * ## Deepview
```
jsaleigh: This appears to be in the wrong section

jsaleigh:
The validation adjustment is the important one, since it appears like this would fail in Server.prototype.getUrl otherwise, as I understand it.

src/6_branch.js
dmitrig01: no, this is right - since the readme is formed only from the comments, and the comment needs start with the function definition, i think we need to do this

src/2_resources.js
dmitrig01: actually no - it looks like it's sending back stringified json - in utils.cleanLinkData, `linkData['data'] = goog.json.serialize(linkData['data'] || {});`

jsaleigh: Ah, cool, in that case...

jsaleigh:
:+1:

dmitrig01:
pushed up one small change, which is that open_app was not defaulting to true properly. fixed now.

dmitrig01:
lemme see why tests are failing

dmitrig01:
ok we're good to go

PR #212, opened 2015-10-19
sendSMS accepts app_short_identifier/ links

Kirkkt:
@dmitrig01 @qinweigong


src/6_branch.js
```diff
@@ -1009,7 +1009,7 @@ Branch.prototype['sendSMS'] = wrap(
 					self._api(
 						resources.linkClick,
 						{
-							"link_url": 'l/' + url.split('/').pop(),
+							"link_url": url.substring(url.lastIndexOf(url.match(/\/\w\//)[0])),
```
dmitrig01: not sure if this does this - but we need to get two url fragments - so we want to support the following usecases:

- `http://bnc.lt/l/abcdef` -> `l/abcdef`
- `http://abc.com/l/abcdef` -> `l/abcdef`
- `http://bnc.lt/m/abcdef` -> `m/abcdef`
- `http://abc.com/m/abcdef` -> `m/abcdef`
- `http://bnc.lt/XYZ/abcdef` -> `XYZ/abcdef`
- `http://abc.com/XYZ/abcdef` -> `XYZ/abcdef`

Kirkkt:
@dmitrig01 @qinweigong PTAL

Right now, if the api pass back app_short_url/link_id it will break web sdk.

So the first step is to review this PR, and then make a web sdk release.

Kirkkt:
oops, I forgot to set the right github label.

@dmitrig01 @qinweigong PTAL

src/1_utils.js
```diff
@@ -263,3 +263,20 @@ utils.base64encode = function(input) {
 	}
 	return output;
 };
+
+/**
+ * Extract the deeplink path from urls in the forms of:
+ * - "protocol://domain.name/deeplink.path
+ * - "domain.name/deeplink.path"
```
qinweigong: do you mean the return value should be "deeplink.path" ?

Kirkkt: it's just a place holder name, unrelated to our internal concept of `$deepview_path`. This function is just a string parser.

Serious, what is the name of the part of the url after the domain name? Like, for example, the url is https://github.com/BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK/pull/212#discussion_r42532884 , what words do we use to describe the `BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK/pull/212#discussion_r42532884` part?

dmitrig01: i think qinwei is correct, though, if you're going for the logic in that comment, you would only return `deeplink.path` - i think that's what this function does, but the comment doesn't make that clear

that part of the url is usually called the "path"

Kirkkt: oh sorry I misunderstood your comment, these two lines are the "accepted" forms of input. Yeah, I should add output form as well. Good point.

Kirkkt: FWIW, the code logic is correct, but I agree the comment I wrote was misleading

Kirkkt:
@dmitrig01 @qinweigong PTAL :)

src/6_branch.js
```diff
@@ -1013,7 +1013,7 @@ Branch.prototype['sendSMS'] = wrap(
 					self._api(
 						resources.linkClick,
 						{
-							"link_url": 'l/' + url.split('/').pop(),
```
qinweigong: sorry, i am not familiar with the context here (first time review web sdk code), but what do we do with the link domain here?

Kirkkt: ah, happy to explain!

as you have seen above, the "before" logic was, oh what a link gets returned from the api backend? TREAT IT AS AN L LINK!

This is very inconvenient to what you are trying to do in api backend, because now, when the v1/url returns `Z2Qa/a1b2c3d4`, the innocent old Web SDK will try to hit `bnc.lt/l//a1b2c3d4` which will make the link service return an equivalent of 404.

So what I did here is to hit `l/...` for l links, `Z2Qa/...` for Z2Qa/ links, `AB12/...` for AB12 links. Or if you will, to make it more general.

Kirkkt: To answer your original question more literally, the domain is stored separately and we don't need to worry about it here

qinweigong: ahh, ok, thanks for the explain.

qinweigong:
:+1:

dmitrig01:
:+1:

PR #215, opened 2015-10-24
Fixing the testing scripts

jsaleigh:


jsaleigh:
@Kirkkt Please have a look, see if this seems correct

Kirkkt:
awesome! Thanks, John!

One thing real quick: make sure to run `make` under the web sdk root directory. this will update the dist/ folder according to your changes in the src/ folder. without running `make`, changed made in src folder won't be effective.

test/6_branch.js
```diff
@@ -507,7 +507,7 @@ describe('Branch', function() {
 			var assert = testUtils.plan(3, done);
 			branch.logout(function(err) {
 				assert(true);
```
Kirkkt: remove. I suspect this will give you a red in the test.html test.

2 lines above it said `var assert = testUtils.plan(3, done);` what this means is that we only plan to do three assertions. So the assertion below is not even run. When you remove this line, the below test will run and it should fail :(

We need better testing framework. One thing I always wanted to do but never got around to is to make it yell when there are actually more assertions than planned (it yells on fewer, however).

jsaleigh:
@Kirkkt new version up with the extra assert removed, keeping the one that was uncommented instead, and with the dist folder added

src/6_branch.js
```diff
@@ -67,6 +67,10 @@ var wrap = function(parameters, func, init) {
 			 * @type {function(?Error,?): undefined}
 			 */
 			var done = function(err, data) {
+				if (err) {
```
Kirkkt: this will make line 74 useless right?

test/6_branch.js
```diff
@@ -526,7 +525,7 @@ describe('Branch', function() {
 				var assert = testUtils.plan(6, done);
 				branch.logout(function(err) {
 					assert(true);
```
Kirkkt: remove

test/6_branch.js
```diff
@@ -664,7 +663,7 @@ describe('Branch', function() {
 				linkData,
 				function(err) {
 					assert(true);
```
Kirkkt: remove

test/6_branch.js
```diff
@@ -685,7 +684,7 @@ describe('Branch', function() {
...
```
Kirkkt: remove

test/6_branch.js
```diff
@@ -902,7 +901,7 @@ describe('Branch', function() {
 			var assert = testUtils.plan(3, done);
 			branch.redeem(1, 'testbucket', function(err) {
 				assert(true);
```
Kirkkt: remove

src/6_branch.js
jsaleigh: Line 74 is still useful, I think. The problem was that if there was an error but no callback, the error was being thrown, which broke the callback chain in the task queue since nothing was calling next().  This meant that the next task added to the queue would never be fired, since the first task wasn't actually removed.

jsaleigh:
@Kirkkt Oops yeah I missed a few asserts...please have a look when you have time

src/6_branch.js
Kirkkt: in that case would it be better to put the queue cleaner inside the next if statement before the throw instead?

Kirkkt: wait... could you do a quick manual test for me:

say we have two branch.something() functions, both don't have callback, they can be the same function.

Let's say I do this:

```
branch.somefunction(someinput1);
branch.somefunction(someinput2);
```

if I run them like this they should queue up, right? and say the first one triggers an err, the second one theoretically should still be called, because it should execute regardless of the previous items in the queue.

Would your change sustain this behavior?

dmitrig01: good catch! perhaps we should call next() instead?

dmitrig01: i mean, before throwing

jsaleigh: @dmitrig01 I was thinking of that, maybe `setTimeout(next)` so that the error on the previous task could be handled before the next task fires, and then remove the call to `next()` on line 83. My reasoning for choosing the current route was that there didn't seem to be a way for a consumer to cancel subsequent tasks on an error, so this would basically force the issue. I'm on the fence about that, what are your thoughts?

dmitrig01: hm, I think that's fine. If they really care, they can nest callbacks

dmitrig01: but the setTimeout idea is great

jsaleigh: @Kirkkt Yup I tried that, what happens if there are no callbacks is that this thread stops execution with a `throw` so never gets to the `next()` call later on. Since there is one dead task at the head of the queue at that point, any other tasks added later would not be fired off.

jsaleigh: As for putting the reset within the next if statement, what I saw is that some of the tests fired off asserts in callbacks, so the same thing happened.

src/6_branch.js
```diff
@@ -693,7 +696,7 @@ Branch.prototype['logout'] = wrap(callback_params.CALLBACK_ERR, function(done) {
 		self.identity = data['identity'];
 		session.update(self._storage, data);
 
-		done(err);
+		done(null);
```
jsaleigh: Quick note on this, there is an earlier check for err that would call `done(err)`. Since err will be undefined or null at this point, I'm forcing it to null for consistency

jsaleigh:
Changes made to revive the queue, ptal

src/6_branch.js
dmitrig01: sounds good

src/6_branch.js
Kirkkt: Ah, thanks letting me know!

src/6_branch.js
```diff
@@ -76,7 +80,6 @@ var wrap = function(parameters, func, init) {
 				else if (parameters === callback_params.CALLBACK_ERR_DATA) {
 					callback(err, data);
 				}
-				next();
```
Kirkkt: By removing this, if there is no error, next() won't get called. Is it intentional?

jsaleigh: Oh wow, good catch.  That could have been bad, thanks Kirk!

Kirkkt: one more thing, if you resume this `next();` here, make sure it doesn't get called twice for the case where we have err and the parameters is not NO_CALLBACK.

This is, btw, why I originally recommended putting your logic in the same if statement as where the throw is located. But let me know if this doesn't make sense

jsaleigh:
@Kirkkt Changed the queue continuation to handle errors better, PTAL when you can

src/6_branch.js
```diff
@@ -67,16 +67,23 @@ var wrap = function(parameters, func, init) {
...
+					if (err && parameters === callback_params.NO_CALLBACK) {
+						throw err;
+					}
+					else if (parameters === callback_params.CALLBACK_ERR) {
+						callback(err);
+					}
+					else if (parameters === callback_params.CALLBACK_ERR_DATA) {
+						callback(err, data);
+					}
+				} catch (err) {
```
dmitrig01: not sure why the linter isn't catching, but cna you put the catch on a new line?

dmitrig01: same with the finally - great ifnd on how to do this

dmitrig01: although actually, you should just be able to omit the catch altogether

src/6_branch.js
```diff
@@ -67,16 +67,23 @@ var wrap = function(parameters, func, init) {
...
+					// the callback functions may or may not throw an error...
 					throw err;
```
Kirkkt: the point of `throw` is to yell and halt the execution because presumably from that point on further execution will be pointless; but here we don't want to halt.
Therefore can we get away with simply `console.error(error message here)`? This try-throw-finally block smells.

src/6_branch.js
jsaleigh: Good point, something in the back of my brain was telling me that finally wouldn't work without it, but yeah catch is optional

src/6_branch.js
jsaleigh: The key is that `next` needs to be called in all cases, but we still also need to bubble errors up to the call stack.  Without that `catch` clause does it feel better to you?

jsaleigh:
Catch removed, finally on the new line...yeah, strange that the linter didn't catch it, it has in the past I thought...

dmitrig01:
great. looks good to me :+1:

src/6_branch.js
```diff
@@ -67,16 +67,21 @@ var wrap = function(parameters, func, init) {
...
 			 */
 			var done = function(err, data) {
-				if (err && parameters === callback_params.NO_CALLBACK) {
-					throw err;
-				}
-				else if (parameters === callback_params.CALLBACK_ERR) {
-					callback(err);
+				try {
+					if (err && parameters === callback_params.NO_CALLBACK) {
+						throw err;
```
Kirkkt: Could you answer my previous question: why don't we just console.error here?

jsaleigh: The expectation for several calls is to throw an error if the Branch SDK has not been initialized. Further down from the `done` method, there are conditionals which check for various states: init pending, init failed, or not inited at all.  Each of those states passes the error to the `done` method, so that it can either be passed to a callback function if one exists, or thrown on line 72 if there is no callback.  The default callback for some methods defined in 6_branch.js also throws the error passed to it, so in lines 75 and 78, there is potential for an error being thrown.  But in all cases, the SDK User expects those errors in the specific cases where initialization fails.

Kirkkt:
:+1:

PR #221, opened 2015-10-29
Allow "identity" to be returned from branch.init()

jsaleigh:
@Kirkkt  This is to allow the identity to be returned in branch.init() for Cordova.  Is this the right way to do this, and should this data be returned there?

### DIFF
--- a/src/6_branch.js
+++ b/src/6_branch.js
@@ -332,6 +332,8 @@ Branch.prototype['init'] = wrap(
 			}
 			if (data['identity_id']) {
 				self.identity_id = data['identity_id'].toString();
+				// to make acceptable for utils.whiteListSessionData
+				data['identity'] = self.identity_id;
 			}
 			if (data['link']) {
 				self.sessionLink = data['link'];


Kirkkt:
would you kindly provide some background on why we need identity in cordova?

chanian:
This was for SDK parity, and a partner request

Sarkar:
nitpick - we return it as 'identity_id' (and not 'identity') inside the other SDKs. 'identity' is reserved for developer identities, and i see that the key from the response is identity_id. 

Just a bit of feedback

Kirkkt:
but `identity` and `identity_id` are two distinct concepts:

`identity` is set by the developer to whatever string they desire
`identity_id` is what branch generates for them: see `branch.setIdentity()` for more details.

Sarkar:
@Kirkkt - yes, exactly. we're returning the branch generated 'identity_id'. however, this code sets the branch generated 'identity_id' to the key 'identity', whereas I think, in order to maintain feature parity w/ the native SDKs, it should be set to the key 'identity_id' (since that's what it is - not the developer set identity)

jsaleigh:
In that case, my change is bad, I need to revert it.  Plus, it sounds like there's a separate 'identity' field that I need to find, and figure out why it isn't available.  Thanks!

jsaleigh:
@Sarkar The `identity_id` isn't being returned from `branch.init()` either, should that field be added as well?  And if so, @Kirkkt would it make sense for me to add that to `utils.whiteListSessionData()`?

Kirkkt:
so it seems to me `identity_id` is some UUID type of thing doesn't make much sense to human
while `identity` is something that, as a developer, I could read, like "john.smith@gmail.com", I think we should refrain from returning `identity_id` (e.g. 4310574) unless there is a good justification.

if branch.init() is not returning `identity` (e.g. john.smith@gmail.com) we have a bug. Fixing this bug should be the scope of this PR; if we need to expose `identity_id`, that is a somewhat unrelated discussion.

Let me know if you disagree.

Kirkkt:
if it's a new user that we don't recognize, `identity` being `null` is normal - my understanding.

PR #228, opened 2015-11-03
redirect for those who bookmarked our web guide doc

Kirkkt:
@aaustin


Kirkkt:
ping @aaustin 

even though by reviving WEB_GUIDE from the dead last Thursday or so, the links in WEB_GUIDE are out of sync with the latest docs.

aaustin:
So - that doc that Jed wrote at "https://dev.branch.io/recipes/quickstart_guide/web/" needs quite a bit of work. I hate to do this but can we not point people at that page until we get this fixed?

PR #233, opened 2015-11-06
Banner text update

jsaleigh:
The `hasApp` call was not always made, especially on fresh installs or when browser storage fails


jsaleigh:
@Kirkkt PTAL

Kirkkt:
:+1:  nice!

PR #242, opened 2015-11-19
fix banner

dmitrig01:
@jsaleigh 

looks like the minifier shat on our variable names


chanian:
:+1:

src/5_banner.js
```diff
@@ -125,13 +125,10 @@ banner = function(branch, options, linkData, storage) {
...
-				if (err) {
-					throw err;
-				}
-				doc.getElementById('branch-mobile-action').onclick =
-					'branch.deepviewCta()';
-			});
+			branch['deepview'](linkData, options);
+			doc.getElementById('branch-mobile-action').onclick = function() {
+				branch['deepviewCta']();
+			};
```
jsaleigh: could this create a race condition? it seems possible (though unlikely) that the call to branch['deepview'] could still be waiting for the ajax request to finish, when the element is clicked on

dmitrig01: deepviewCta puts it on the queue, which will wait for deepview to complete

jsaleigh:
:+1:

PR #243, opened 2015-11-23
Revert to POJO storage when localStorage is not available

### LINKED ISSUE #210: SDK breaks when sessionStorage/localStorage are disabled
When I disable localStorage and sessionStorage in Chrome (Settings Tab-> Advanced -> Content Settings -> Block sites from setting any data), the SDK triggers a security error. A similar error will happen in any browser with localStorage disabled. For example, when I open the example page linked in the readme, I see this message:

image

This could be solved by using a `utils.localStorage` object that is set to window.localStorage when allowed, and uses a polyfill when not allowed. I'd be happy to submit a PR.

elsbree:
Fixes #210


jsaleigh:
Closing out as resolved

PR #244, opened 2015-11-24
Gracefully handle local and session storage unavailability

jsaleigh:


src/2_storage.js
```diff
@@ -61,7 +61,17 @@ var retrieveValue = function(value) {
 };
 
 var webStorage = function(perm) {
-	var storageMethod = perm ? localStorage : sessionStorage;
+	var storageMethod;
+	try {
+		storageMethod = perm ? localStorage : sessionStorage;
```
dmitrig01: so is the issue that referencing localStorage or sessionStorage is returning undefined?

jsaleigh: No, an attempt to access either localStorage or sessionStorage when storage is disabled is throwing an Error.  Even just trying to assign either to storageMethod in this case is throwing the error.

dmitrig01:
cool, :+1:

PR #246, opened 2015-12-02
use deepview for banner in all scenarios, even if there is a referring link

dmitrig01:
@jsaleigh 

deepview includes handling for referring link, we should definitely not be putting it behind the banner button, otherwise it'll just redirect to the page you're already on


jsaleigh:
Makes sense :+1:

jsaleigh:
@dmitrig01 the conflicts look like just build files, let me know if you would like me to fix that and merge

dmitrig01:
yes please!!

dmitrig01:
let's add a changelog entry too - how about "fixes banner call to use deepview logic even if there is a referring link"

PR #247, opened 2015-12-03
Bug fix for issues#240

sweetlikepete:
Added a check to ensure that the cached $desktop_url is equal to the current url before returning a cache hit. I tested to make sure that only subsequent loads of the same page hit the cache and new pages always hit /v1/open. I left in the all the other checks just to be safe.


jsaleigh:
Hey @sweetlikepete, thanks for this fix!  I'm going to get it into a function in a moment, so that the variables are only set / created in the web build.  Please let me know if you see any issues with it :)

src/6_branch.js
```diff
@@ -376,6 +376,17 @@ Branch.prototype['init'] = wrap(
 			(url ? utils.getParamValue(url) : null);
 		var freshInstall = !sessionData || !sessionData['identity_id'];
 
+		function currentDesktopUrlMatchesSession() {
+			var currentDesktopUrl = window.location.protocol + "//" + window.location.host + window.location.pathname;
```
dmitrig01: protocol will have a :?

dmitrig01: just checked. guess so. weird

sweetlikepete: iknowright

PR #248, opened 2015-12-03
Banner design tweaks and extensions

sweetlikepete:
This pull contains some of the proposed tweaks and extensions to the banner. 

https://app.asana.com/0/48043688138853/66927422477320/f

There are a few new options in branch.banner();

'theme' : 'light' or 'dark'  - light looks pretty much the same as before, dark is new
'rating' : 0.5 to 5 - will show a star rating from 0.5 - 5 
'reviewCount' : 0 to n - will show how many people have submitted reviews
'buttonBorderColor' : css color

Although these weren't asked for explicitly, I figured they made sense in combination with buttonBorderColor.

'buttonBackgroundColor' : css color
'buttonFontColor' : css color
'buttonBorderColorHover' : css color
'buttonBackgroundColorHover' : css color
'buttonFontColorHover' : css color

I did my best to make sure the styles work across all the permutations of devices we support, however I don't have all the devices themselves to test on, so I tested with a user agent switcher only. It should probably be tested by somebody more familiar with the functionality than myself.


src/6_branch.js
```diff
@@ -1782,6 +1782,19 @@ if (WEB_BUILD) {
 			icon: options['icon'] || '',
 			title: options['title'] || '',
 			description: options['description'] || '',
+			reviewCount: (
+				typeof options['reviewCount'] === 'number' &&
+				options['reviewCount'] > 0 // force greater than 0
+			) ?
+				Math.floor(options['reviewCount']) : // force no decimal
+				null,
```
jsaleigh: Minor nit, but this could go on the same line:
) ? Math.floor(options['reviewCount']) : null

jsaleigh:
:+1:

PR #254, opened 2015-12-14
Moved change from branch amd_fix to this branch

sweetlikepete:
Old branch was amd_fix_without_code_style_changes. This branch contains the change without any of the code styling changes.


wlindemann:
@sweetlikepete I left a note in Asana, but if you can work on testing this - we don't have anyone here who is ready to go on that.

sweetlikepete:
No problem. My understanding is that the library should load and function correctly through requirejs.

I've only read about it - haven't used requirejs before in production. I know the pattern from using nodejs. Will set it up and test that it works correctly in browser.

dmitrig01:
requirejs == AMD == using define() functions == loading code async

commonjs == using require() functions == loading code synchronously == what node uses and what browserify uses

dmitrig01:
umd (what we have in the web sdk) is a wrapper that supports both

dmitrig01:
(ideally)

jsaleigh:
:+1:

PR #258, opened 2015-12-20
Added an event for didDeepviewCTA

sweetlikepete:
This was asked for @janpaul123 here https://github.com/BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK/issues/255#issuecomment-165923976 - it's not the exact kind of name that was asked for, but I think it makes sense in this context in case something else calls deepviewCTA at some point in the future.


janpaul123:
Fantastic! Thanks for this.

jsaleigh:
:+1:

PR #259, opened 2015-12-21
Support App Links on Android

aaustin:
@jsaleigh It’s so confusing that this is in the web SDK.. argh


jsaleigh:
:+1:  Yeah I agree. We should be able to code this such a way that Cordova can just override or augment functionality, rather than ifdefs in the Web SDK.

PR #264, opened 2016-01-07
The referring link data was not held when qsps were removed

jsaleigh:


src/6_branch.js
```diff
@@ -472,12 +472,14 @@ Branch.prototype['init'] = wrap(
...
 				sessionData['session_id'] &&
-				currentDesktopUrlMatchesSession() &&
-				(utils.processReferringLink(link_identifier) === sessionData['referring_link'] ||
-				link_identifier === sessionData['click_id'])) {
-			attachVisibilityEvent();
-			checkHasApp(sessionData, finishInit);
-			return;
+				currentDesktopUrlMatchesSession()) {
+
+			if ((!link_identifier && sessionData['referring_link']) ||
```
jsaleigh: The "link_identifier" value in this case is either the '_branch_match_id' or 'r' qsp

jsaleigh:
@dmitrig01 If you would be so kind...

jsaleigh:
The tests are failing...digging in to that first

test/6_branch.js
```diff
@@ -346,7 +346,12 @@ describe('Branch', function() {
 				{
 					session_id: session_id,
 					browser_fingerprint_id: browser_fingerprint_id,
-					identity_id: identity_id
+					identity_id: identity_id,
+					data: JSON.stringify({
+						'$desktop_url': window.location.protocol + "//" +
```
jsaleigh: needed for call to currentDesktopUrlMatchesSession

package.json
```diff
@@ -18,10 +18,12 @@
 		"url": "https://github.com/BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK/issues"
 	},
 	"homepage": "https://github.com/BranchMetrics/Smart-App-Banner-Deep-Linking-Web-SDK",
+	"peerDependencies": {
```
jsaleigh: CircleCI needs grunt as a peer dependency, else it will throw errors during the tests

PR #265, opened 2016-01-12
webpack requires the module definition to be a string

jsaleigh:


jsaleigh:
@sweetlikepete I had to reverse one of your recent changes (that was from the `amd_fix` branch originally), but I need to make sure this won't break something else...can you take a quick look?

PR #268, opened 2016-01-12
Multi path fix

jsaleigh:
@aaustin This reverses a recent change that is causing issues for partners


aaustin:
I have no idea the purpose of that for that if-clause. :+1:

PR #270, opened 2016-01-14
banner hiding data should be in local, not session, storage

jsaleigh:


dmitrig01:
it should only be in local if forgetHide is true right? so we'd need to read from both places?

chanian:
Pending @dmitrig01's feedback, looks good to me. :+1:

dmitrig01:
:+1:

PR #273, opened 2016-01-25
update minified branch source to v1.8.8

mrblueblue:


aaustin:
Oops ! Thanks @mrblueblue

PR #274, opened 2016-01-27
Same session fix

jsaleigh:
Force /v1/open call when there is a link identifier
Added failing test to handle same-domain inits
Fixed some existing tests

### DIFF
--- a/src/6_branch.js
+++ b/src/6_branch.js
@@ -458,16 +458,10 @@ Branch.prototype['init'] = wrap(
 			}
 		};
 
-		if (WEB_BUILD &&
-				sessionData &&
-				sessionData['session_id']) {
-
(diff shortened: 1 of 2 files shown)


jsaleigh:
The only change is that in every case where ?_branch_match_id or #r is present, a call to /v1/open will be made.  This is to handle two different issues:

1) When a partner has a multi-page flow, they should be able to maintain referring link data

2) When two referring links are used in the same session (Facebook app's webview, for example), the second link should not use the cached link data.

jsaleigh:
@aaustin @dmitrig01 

We need to be sure that this does not damage the currently intended functionality.

aaustin:
This seems like it should cover all cases :+1:

PR #275, opened 2016-01-28
Remove grunt-cli as dependency

floriangosse:
grunt-cli is not needed for the use of the sdk and throws warnings if you try to install it.


jsaleigh:
This change causes CircleCI to fail while connecting to Sauce Labs.

PR #278, opened 2016-01-28
remove grunt-cli from peer dependencies

floriangosse:


jsaleigh:
:+1: This works when running by hand on the server, so I'm merging it in while we figure out the Sauce Labs issues

floriangosse:
Thank you.

PR #279, opened 2016-02-01
Fixed issue with ipad showing banner even though showiOS is false

sweetlikepete:
…false


src/3_banner_utils.js
```diff
@@ -165,7 +165,7 @@ banner_utils.shouldAppend = function(storage, options) {
 		(
 			(options.showDesktop && !utils.mobileUserAgent()) ||
 			(options.showAndroid && utils.mobileUserAgent() === 'android') ||
-			(options.showiPad && utils.mobileUserAgent() === 'ipad') ||
+			(options.showiPad && (utils.mobileUserAgent() === 'ipad' && options.showiOS)) ||
```
aaustin: This handles the case where showiPad is false and showiOS is true, which was broken before. However, it breaks the case where showIpad is true and showiOS is false.

PR #280, opened 2016-02-05
Updated README.md

ankit-saxena:
removed reference to funnels


jsaleigh:
@ankit-saxena I'm afraid these changes would get wiped out on a build, since README.md is auto-generated.  I believe the file you'll want to change is src/6_branch.js, around line 758.

aaustin:
Updated this on your behalf

PR #282, opened 2016-02-13
Only use the closure compiler's JSON.parse as a fallback...otherwise …

jsaleigh:
…'eval' is used instead by default, which can cause Content Security Policy issues.


jsaleigh:
Now that I look at it, this could fail.  goog.json.parse has exception handling, so a straight replacement loses that while it gets rid of eval.

docs/web/3_branch_web.md
```diff
@@ -464,7 +464,7 @@ branch.deepview(data, option, function(err) {
     if (err) {
         throw err;
     }
-    $('a.deepview-cta').click(branch.deepviewCta);
+    ${'a.deepview-cta').click(branch.deepviewCta);
```
chanian: Typo here?

chanian:
:+1: contingent on fixing typo in docs

docs/web/3_branch_web.md
jsaleigh: Wow where did that come from...

PR #292, opened 2016-03-04
Don't check for apps on desktop

jsaleigh:
@dmitrig01


dmitrig01:
hope this doesn't bite us in the ass :+1:

jsaleigh:
i'll have to update the unit tests to handle this change

PR #300, opened 2016-03-11
Revert "Merge pull request #299 from BranchMetrics/cleanup-revert"

jsaleigh:
This is to reapply the cleanup merge that needed to be temporarily reverted


jsaleigh:
This is the same code that was reviewed previously, so I'm merging it in now.  This will need QA before a deployment.

PR #301, opened 2016-03-14
don't default append_deeplink_path to true

derrickstaten:
@jsaleigh


jsaleigh:
Looks valid, I have to wait for the current `master` validation before thumbing

PR #305, opened 2016-03-21
Automatically populate the Open Graph metadata used in Links...

jsaleigh:
...but only if they have not already been set.

See: http://ogp.me/

This code will scrape the page for meta tags containing title, description, and image information when that information was not already set in calls to branch.banner(), branch.link(), and branch.sendSMS().

@jeanmw you're the best one to PR this, if you could.

@scott-branchio testing this would involve setting meta tags in your testing pages.


example.html
```diff
@@ -2,6 +2,9 @@
 <html>
 <head>
 	<meta name="viewport" content="width=device-width, initial-scale=1">
+	<meta content="http://branch.io/img/logo_icon_black.png" property="og:image" />
+	<meta content="Branch Metrics Web SDK Example App" property="og:title" />
+	<meta content="A basic example to demonstrate some of the ways that the Web SDK can be used" property="og:description" />
 	<title>Branch Metrics Web SDK Example App</title>
```
jsaleigh: These extra tags are to show how the data can be pulled from the head if they are left out of calls

jsaleigh:
@rubinsingh @jeanmw this one is much more interesting...the intention is to scrape the head section for og tags to use, when that information was not already specified by the developer

jsaleigh:
Hmm actually I'll need to look at it, the failing checks are not ignorable this time around :(

src/1_utils.js
```diff
@@ -105,20 +105,66 @@ utils.whiteListSessionData = function(data) {
...
+/**
+ * Abstract away the window.location for better testing
+ */
+utils.getWindowLocation = function() {
+	return String(window.location);
+};
+
 utils.cleanLinkData = function(linkData) {
 	linkData['source'] = 'web-sdk';
-	if (linkData['data'] && linkData['data']['$desktop_url'] !== undefined) {
```
jeanmw: why did we remove the check for if desktop_url is undefined?

test/6_branch.js
```diff
@@ -635,7 +635,9 @@ describe('Branch', function() {
 					'$desktop_url': 'https://cdn.branch.io/example.html',
 					'$og_title': 'Branch Metrics',
 					'$og_description': 'Branch Metrics',
-					'$og_image_url': 'http://branch.io/img/logo_icon_white.png'
+					'$og_image_url': 'http://branch.io/img/logo_icon_white.png',
+					'$canonical_url': 'https://cdn.branch.io/example.html',
```
jeanmw: looks good!

jeanmw:
@jsaleigh It's probably nothing but I had one question on why you removed a check for null $desktop_url (see comment) but other than that LGTM

src/1_utils.js
jsaleigh: The check for desktop_url used to be the only one, but once we started adding more types, I went for a refactor that checked desktop_url on line 152 of the newer code.  And then, in that case, I am specifically looking for a 'string' type, instead of the absence of the value.

jsaleigh:
@jeanmw I've added an answer, if it sounds reasonable to you I'll merge

src/1_utils.js
jeanmw: Oh I see what you're doing! Cool LGTM :)

jeanmw:
@jsaleigh  cool I see what you're doing, looks good to me!

PR #306, opened 2016-03-22
Clean up after JSONP calls

jsaleigh:
- Remove script tags after JSONP calls
- Remove JSONP callbacks after they are used
- Prevent a few caught errors from happening in the first place


dist/build.js
```diff
@@ -794,8 +803,11 @@ utils.mobileUserAgent = function() {
 };
 utils.getParamValue = function(a) {
   try {
-    return utils.getLocationSearch().substring(1).match(new RegExp(a + "=([^&]*)"))[1];
-  } catch (b) {
```
jeanmw: looks good, seems like this is functionally equivalent but cleaner

jeanmw:
:+1: looks good to me, looks like a lot of this is now just much more nicely refactored :)

PR #312, opened 2016-03-28
Set branch view data

jsaleigh:
This PR is meant to persist user defined data for reuse, ensure that subsequent attempts to set the data overwrites previous attempts, and paves the way for Branch Views

@rubinsingh @jeanmw, when you have a chance please


src/6_branch.js
```diff
@@ -1654,11 +1657,16 @@ Branch.prototype['banner'] = wrap(callback_params.NO_CALLBACK, function(done, op
 		bannerOptions.showKindle = options['showMobile'];
 	}
 
+	this.setBranchViewData(data);
 	this.closeBannerPointer = banner(this, bannerOptions, data, this._storage);
 	done();
 });
 
-Branch.prototype['closeBanner'] = wrap(0, function(done) {
```
jeanmw: makes sense - will the behavior/ availability of 'closeBanner' be otherwise affected at all in this merge?

dist/build.js
```diff
@@ -1692,7 +1702,10 @@ Branch.prototype.banner = wrap(callback_params.NO_CALLBACK, function(a, b, c) {
   this.closeBannerPointer = banner(this, d, c, this._storage);
   a();
 });
-Branch.prototype.closeBanner = wrap(0, function(a) {
+Branch.prototype.setBranchViewData = wrap(callback_params.NO_CALLBACK, function(a, b) {
+  this._bvData = goog.object.clone(b || {});
+});
+Branch.prototype.closeBanner = wrap(callback_params.NO_CALLBACK, function(a) {
```
jsaleigh: The change from `wrap(0` to `wrap(callback_params.NO_CALLBACK` is not entirely necessary here, but makes for more consistent code

jeanmw:
LGTM @jsaleigh , added one comment above more out of curiosity (I think the answer to my question is no), and tested it out, seems good, let's see what circleci says  :)

dist/build.js
jeanmw: ah ok cool

src/6_branch.js
jsaleigh: It should be exactly the same

jsaleigh:
Rerunning the CircleCI tests...digging in to the SauceLabs URLs show passes, but CircleCI is reporting the same as fails.

PR #313, opened 2016-03-28
attempt fix to set has_app using correct session

jeanmw:
@jsaleigh Not sure if this is right, but may fix the checkHasApp function - let me know what you think or if this needs further modifications


jsaleigh:
Looks good to me :+1:

jeanmw:
manually ran circleci tests but they all time out :/

jsaleigh:
@jeanmw let's hold off on this one before a merge...I was thinking about a potential issue, want to talk it out before we go forward

jsaleigh:
@jeanmw @scott-branchio the backend changes are in place to test this...let's get it out once it passes all tests

scott-branchio:
@jsaleigh @jeanmw I am unable to get **has_app** to come back with "true" using iPhone Safari (9.3.1) browser after trying for about an hour.  When using Chrome on same device it works, although there is a side-effect of loading the smart banner and the BMF app is directly launched instead of waiting for me to tap on the download button @derrickstaten might be interested in that fact.

It's unclear what the expected results are when testing as I have never seen this work consistently.  Here the web page where I have BMF links to smart banners - http://qa:tesT123@qaauto.branch.io/qa/test_cases/links.html

Towards the middle of the page there is a **QA Branchster Test Cases** section, I tap on any of those 4 links to perform the testing - using Safari devloper and Charles Proxy.  It seems there is some disconnect with the browser_fingerprint_id and the device_fingerprint_id that makes **has_app** always return false when using iOS Safari.

I will need some hands on help troubleshooting this issue to get an consistent expected results - this way I can document a concrete test case for future testing.

scott-branchio:
@jeanmw @jsaleigh is anything happening here with this PR?  If there is a backend problem it would be great to identify it and get that fixed?

scott-branchio:
@jsaleigh @jeanmw I've completed testing this on my iPhone 9.3.1 using Safari and the example.html page and verified that the BFP is either retrieved from the new "_t" parameter or cookie (if it exists). 👍

Issue #319, opened 2016-04-20
TypeError: object does not support the 'remove' property or method .


jamesfzhang:
This error occurs for users on IE11. I see variations from all different languages. The trace points to `
File https://cdn.branch.io/branch-latest.min.js line 27 col 492 in Anonymous function`.

jsaleigh:
@jamesfzhang Thanks for the head's up...it appears that this is happening during an attempt to cleanup script tags created during jsonp requests.  We'll get a fix out as soon as possible for it.

jamesfzhang:
Thanks John, appreciate it.

jsaleigh:
@jamesfzhang We've deployed a new Web SDK version to fix this, v2.1.5.  Once the cache clears on `https://cdn.branch.io/branch-latest.min.js`, the build should be available there.

Thanks again for finding this! 😄

PR #320, opened 2016-04-20
Script removal error

jsaleigh:
#319 

When removing script tags after a jsonp request, check that the current browser supports ChildNode.remove()

https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove

@jeanmw @scott-branchio


scott-branchio:
@jsaleigh @jeanmw completed testing IE11 on Win7 env - verified the error using cdn.branch.io/example.html page; tested against updated build.js in QA environment example.html page and did not see the error; also tested QA example.html page using Chrome, Firefox and Safari on Mac and saw no errors. 👍

jsaleigh:
Awesome, thanks @scott-branchio !

PR #324, opened 2016-04-29
sub domain fix for sendsms

jsaleigh:


src/2_resources.js
```diff
@@ -117,7 +117,7 @@ resources._r = {
 };
 
 resources.linkClick = {
-	destination: config.link_service_endpoint,
+	destination: "",
```
jsaleigh: With subdomains, the destination is no longer always bnc.lt

src/3_api.js
```diff
@@ -156,7 +156,7 @@ Server.prototype.getUrl = function(resource, data) {
 	}
 	return {
 		data: this.serializeObject(d, ''),
-		url: url
+		url: url.replace(/^\//, '')
```
jsaleigh: Because `getUrl` joins url fragments with `/`, and because we no longer have a destination for the `linkClick` resource, we need to get rid of the extra slash that can be prepended.

src/6_branch.js
```diff
@@ -839,10 +839,13 @@ Branch.prototype['sendSMS'] = wrap(
 						return done(err);
 					}
 					var url = data['url'];
+					if (!/(bnc.lt\/|app\.link\/)/.test(url)) {
+						url = config.link_service_endpoint + utils.extractDeeplinkPath(url);
+					}
```
jsaleigh: If the url returned from the link service is already on a Branch domain, just use it...otherwise, set it to be a Branch domain (this is just-in-case, and may never be needed)

jsaleigh:
@scott-branchio This should fix the errors with SendSMS caused by the subdomain change...we need to get it deployed as soon as you're able to test it

chanian:
👍 verify with qa that base cases still good

jsaleigh:
Also I need to check why the unit tests are failing.  It is probably because we are forcing a url change, but we need to be sure this didn't actually break anything.

scott-branchio:
@jsaleigh completed testing this PR and created new auto script to test for sending SMS with new app/subdomain 👍

PR #328, opened 2016-05-08
Correct-typo-line600

dwestgate:
@jsaleigh Corrects typo - contins -> contains


jsaleigh:
This wouldn't survive the build process, but I've merged the same changes into #333

PR #329, opened 2016-05-26
we should call preventDefault ...

jsaleigh:
...to stop the cta anchor from forwarding to /#, which breaks single page applications


aaustin:
👍

PR #332, opened 2016-06-06
Add in options to open

Sarkar:
allow options dict to pass through v1/open

@jsaleigh @aaustin


jsaleigh:
Looks right to me 👍

aaustin:
Nice!

PR #333, opened 2016-06-14
Journey div injection

jsaleigh:


src/branch_view.js
```diff
@@ -2,19 +2,42 @@
...
 	if(match) {
 		var src = match[1];
-		html.replace(re,'');
+		html = html.replace(re,'');
+
+		var metadata = safejson.parse(src);
+		if (metadata && metadata.injectorSelector) {
+			var parentTrap = document.querySelector(metadata.injectorSelector);
+			if (parentTrap) {
+				parent = parentTrap;
```
Sarkar: is this going to be problematic if parent isn't present? i think i saw a few cases where there wasn't a parent passed through, and as a result, parent was defaulted to something else

src/branch_view.js
```diff
@@ -23,7 +46,7 @@ function renderHtmlBlob(parent, html) {
 	parent.insertBefore(banner, parent.firstChild);
 
 	banner_utils.addClass(banner, 'branch-banner-is-active');
-	banner.marginTop = '76px';
+	// banner.marginTop = '76px';
```
Sarkar: let's just remove this?

Sarkar:
other than those 2 comments, 👍

src/branch_view.js
jsaleigh: It should be ok, because if there is no parent, it will default to document.body (on line 41).  The old branch.banner() used document.body at all times, so parent should be absent in most cases.

src/branch_view.js
jsaleigh: I was going back-and-forth on it, because I didn't remember why it was there, and left this as a note.  But you're right, I'll remove it.

PR #339, opened 2016-06-28
Always show the banner

nitrag:
Overriding the shouldAppend so that the banner is always shown if specified in the options.


jsaleigh:
@nitrag Thanks for the PR! Did the `disableHide` and `forgetHide` parameters not work for you?

nitrag:
Dang, maybe that was my problem, I guess docs were confusing. That method is working for me. 

My PR is junk anyway, it didn't work.

PR #344, opened 2016-07-07
JOUR-463 don't fire the app by default

jsaleigh:
@zeeshabnc @aaustin 

The reason for adding this line in the first place is lost, and it seems to be causing issues by being in.


zeeshabnc:
👍 Looks good to me.

aaustin:
yep - simple enough 👍

PR #346, opened 2016-07-07
JOUR-449 Journeys button text based on user app installed

yntema:


src/branch_view.js
```diff
@@ -4,8 +4,8 @@ goog.require('utils');
 goog.require('banner_css');
 goog.require('safejson');
 
-function renderHtmlBlob(parent, html, bannerElement) {
-	var re, match;
+function renderHtmlBlob(parent, html, hasApp) {
+	var re, match, ctaText;
```
jsaleigh: In case the metadata script doesn't exist, you might want to set defaults here:

`var re, match, ctaText = hasApp ? 'OPEN' : 'GET'`

yntema: The defaults are currently set on lines 19 and 25. Do you think it makes more sense to set them at the top though?

jsaleigh: Yes, because lines 19 and 25 can only be accessed if the match in line 13 exists.  If the metadata script has not been added to the template (either now or with a dashboard change in the future), the `ctaText` variable will never be set, and the CTA itself will be blank.

yntema: Good call. Updated!

jsaleigh:
👍

PR #355, opened 2016-07-19
jour-479 allow display of preview templates

jsaleigh:
@yntema when you have a chance

### DIFF
--- a/src/6_branch.js
+++ b/src/6_branch.js
@@ -272,6 +272,15 @@ Branch.prototype['init'] = wrap(
 
 		options = (options && typeof options === 'function') ? { } : options;
 
+		var branchViewId = null;
+
+		if (options) {
+			branchViewId = options.branch_view_id || null;
+		}
+		if (!branchViewId) {
+			branchViewId = utils.getParameterByName('_branch_view_id') || null;
+		}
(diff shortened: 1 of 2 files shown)


src/6_branch.js
```diff
@@ -272,6 +272,15 @@ Branch.prototype['init'] = wrap(
 
 		options = (options && typeof options === 'function') ? { } : options;
 
+		var branchViewId = null;
+
+		if (options) {
+			branchViewId = options.branch_view_id || null;
+		}
+		if (!branchViewId) {
+			branchViewId = utils.getParameterByName('_branch_view_id') || null;
```
yntema: Should the underscore be removed from '_branch_view_id' ?

jsaleigh: I did that intentionally, to reduce the chance of collision with partner QSPs.  It might not be needed though.

zeeshabnc: Lets leave the underscore, gives the implication that its internal use

yntema:
👍

src/6_branch.js
yntema: Cool, just wanted to make sure it was intentional

PR #369, opened 2016-08-31
Hosted Deep Link Data in banner and deepview.

TheMarshal99:
Updated the banner() and deepview() calls.
* Will now pull in any hosted deep link data into the data block.
* Items in the data block will override hosted deep link data.

@jsaleigh

### DIFF
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -439,4 +439,21 @@ utils.scrapeOpenGraphContent = function(property, content) {
 	return content;
 };
 
+/**
+ * Search for hosted deep link data on the page, as outlined here https://dev.branch.io/getting-started/hosted-deep-link-data/guide/#adding-metatags-to-your-site
+ */
+utils.scrapeHostedDeepLinkData = function() {
+	var params = {};
+
+	var metas = document.getElementsByTagName("meta");
+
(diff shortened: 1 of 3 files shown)


src/1_utils.js
```diff
@@ -439,4 +439,21 @@ utils.scrapeOpenGraphContent = function(property, content) {
...
+/**
+ * Search for hosted deep link data on the page, as outlined here https://dev.branch.io/getting-started/hosted-deep-link-data/guide/#adding-metatags-to-your-site
+ */
+utils.scrapeHostedDeepLinkData = function() {
+	var params = {};
+
+	var metas = document.getElementsByTagName("meta");
+
+	for (var i = 0; i < metas.length; i++) {
+		if (metas[i].getAttribute("name").startsWith("branch:deeplink:")) {
```
jsaleigh: If there is no 'name' attribute, `getAttribute` will return null, then `startsWith` will crash

jsaleigh: `startsWith` cannot be used:

http://kangax.github.io/compat-table/es6/#test-String.prototype.startsWith

TheMarshal99:
@jsaleigh ^^

yntema:
👍  looks good!

PR #371, opened 2016-09-08
JOUR-590 pass smartbanner options

yntema:
In preparation for migration to journeys. Passing banner option so we have info to build journeys for all apps using smartbanners.

**Review**
@jsaleigh


dist/build.js
```diff
@@ -1300,12 +1300,13 @@ banner_css.css = function(a, b) {
...
   banner_css.ios;
   c += a.customCSS;
-  a.iframe && (c += banner_css.inneriframe, d = document.createElement("style"), d.type = "text/css", d.id = "branch-iframe-css", d.innerHTML = banner_css.iframe + (utils.mobileUserAgent() ? banner_css.iframe_position(a.mobileSticky, a.position) : banner_css.iframe_position(a.desktopSticky, a.position)), document.head.appendChild(d));
+  a.iframe && (c += banner_css.inneriframe, d = document.createElement("style"), d.type = "text/css", d.id = "branch-iframe-css", d.innerHTML = banner_css.iframe + (utils.mobileUserAgent() ? banner_css.iframe_position(a.mobileSticky, a.position) : banner_css.iframe_position(a.desktopSticky, a.position)), (document.head || document.getElementsByTagName("head")[0]).appendChild(d));
```
zeeshabnc: What if head doesn't exist?

yntema:
@zeeshabnc I'm not sure where that came from. Must have been some other stuff when I ran `make`. Let me take another look

yntema:
I'm going to close this out and create a new one. It looks like some of the other changes I was working on were in the build files.

yntema:
@zeeshabnc just realized, those were changes related to the last pull request which didn't include the build files. Should be all good. The change was put in because in some cases document.head was not found, so it is searched for in a different way. the head is always there on html documents

zeeshabnc:
👍

PR #372, opened 2016-09-14
JOUR-638 Pass initial_referrer in v1/open and v1/event requests

yntema:
Passing initial referrer to enhance audience rules based on the previous page a user came from.


jsaleigh:
👍

PR #378, opened 2016-09-16
WEBSDK-13 stop throwing error if init fails

yntema:


src/6_branch.js
```diff
@@ -66,9 +66,7 @@ var wrap = function(parameters, func, init) {
 		var lastArg = arguments[arguments.length - 1];
 		if (parameters === callback_params.NO_CALLBACK || typeof lastArg !== 'function') {
 			callback = function(err) {
-				if (err) {
-					throw err;
-				}
```
zeeshabnc: Can we add a console.log statement here?

yntema: We can, but it seemed repetitive since we already log the error else where (line 103).

zeeshabnc: ok

jsaleigh:
👍

PR #379, opened 2016-09-19
patch for no data in banner

yntema:


zeeshabnc:
Sweet! Can we add a test too?
👍

PR #384, opened 2016-09-28
INTENG-2844 Make merge and iframe more defensive

yntema:
utils.merge was causing an error if one of the arguments was passing in `null`.
 
smartbanner was occasionally unable to attach the sendSMS call to action. This change makes sure that the iframe has loaded before trying to attach the event listener.


jsaleigh:
👍

PR #385, opened 2016-09-28
Add extractMobileDeeplinkPath util function

yntema:
Needed an additional method to extract mobile deeplink path from app links metadata.

### DIFF
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -424,6 +424,26 @@ utils.extractDeeplinkPath = function(url) {
 };
 
 /**
+ * Extract the path (the part of the url excluding protocol and domain name) from urls in the forms
+ * of:
+ * - "AppName://some/path
+ *
+ * and returns (for the above sample input cases):
+ * - "some/path"
+ *
+ * @param {string} url
+ */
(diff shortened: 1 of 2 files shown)


test/1_utils.js
```diff
@@ -343,4 +343,45 @@ describe('utils', function() {
 			}
 		});
 	});
+	describe('extractMobileDeeplinkPath', function() {
+		it('should return deeplink path mobile scheme url', function() {
+			if (testUtils.go('#test:extractDeeplinkPath')) {
```
jsaleigh: Should this be extractMobileDeeplinkPath?

src/1_utils.js
```diff
@@ -424,6 +424,31 @@ utils.extractDeeplinkPath = function(url) {
...
+ * - "AppName://some/path
+ * - some/path
+ * - /some/path
+ *
+ * and returns (for the above sample input cases):
+ * - "some/path"
+ *
+ * @param {string} url
+ */
+utils.extractMobileDeeplinkPath = function(url) {
```
jsaleigh: Will extractDeeplinkPath be needed after this?

PR #387, opened 2016-09-29
Allow both meta name and property to be in applinks metadata

yntema:
This change accounts for sites that include both a property and name in their app links meta tags.

`<meta name="test" property="test" content="test-content">`

@jsaleigh


jsaleigh:
👍

PR #390, opened 2016-10-13
JOUR-656 - Update Journeys test url behavior

yntema:
Updates to test url behavior:

- no longer show journeys banner on desktop test
- enable CTA and hideBanner buttons
- Add testFlag. Refreshing page will show test banner again even if hide banner is set for a longer period of time
- remove unnecessary handleBranchView() call in v1/open callback. v1/event returns branchViewData, not v1/open

Reviewer: @jsaleigh


jsaleigh:
👍

PR #391, opened 2016-10-13
INTENG-2878 Check localStorage before using

yntema:
https://branch.atlassian.net/browse/INTENG-2878
Issue caused when sdk tries to set something to local storage while user is browsing in Safari private mode. Need to check if localStorage is available before using it.

Also fixed tests that are failing because of hardcoded values.


jsaleigh:
👍

PR #397, opened 2016-11-28
Remove goog.json.parse

yntema:
goog.json.parse uses eval() causing warnings. Getting rid of it since JSON.parse is very well supported.


jsaleigh:
👍  The test failures appear to be just CircleCI being flaky

PR #399, opened 2016-12-08
WEBSDK-30 add creation source to link payload

ahmednawar:
This adds `source` to the link payload that the web sdk sends to the API. This will help identify the link creation_source to be the Web SDK instead of just SDK.
 
@yntema


rubinsingh:
👍

PR #400, opened 2016-12-13
Support half page or partial height banners

yntema:
When declaring banner height in relative units, the content height is also currently using relative units. The problem is that the content height is relative to the banner height and not the screen height. This pr fixes that scenario. This change will only effect interstitial banners that are less than 100vh.

**review: @jsaleigh  @rubinsingh**


rubinsingh:
👍  nice work @yntema

PR #404, opened 2016-12-22
JOUR-723 Allow more than one div insertion for journeys

yntema:
Selecting **ALL** elements with class="branch-journeys-top" and adding appropriate margin to parentElement.

Partners often have multiple fixed elements in their navbar and journeys needs to be able to support this.


jsaleigh:
Once this passes regression tests, lgtm 👍

PR #406, opened 2017-01-03
Remove box-shadow if no content background exists

yntema:
Checks the computed background-color of the .branch-banner-content. If the alpha === 0, we can determine there is no background and thus remove the box-shadow. 

Allows for floating button banner.


aaustin:
@jsaleigh Can you review this for Gerrit?

jsaleigh:
Looks good to me 👍

PR #408, opened 2017-01-05
INTENG-3122 - Allow init to complete when banner dismissed

yntema:
Issue was that if the localStorage said banner should be hidden, the function would return instead of allowing it to finish initializing. 

Moved the dismiss banner logic up so that we could check if the banner should be shown and allow done() to be called if not.


rubinsingh:
@yntema just tested this PR by closing a Journey and then refreshing the page. Initialization worked as expected and I was able to use the other SDK functions 👍

jsaleigh:
The failed CircleCI appears to be a timeout for select browsers, it should be ok.

👍

PR #413, opened 2017-01-11
Fix floating button support with parseFloat

yntema:
Need to use parseFloat to handle floating point alpha values.


rubinsingh:
looks great 👍

PR #414, opened 2017-01-16
JOUR-720 add support for no_journeys option in init

yntema:
Partners are asking for a way to be able to prevent a journey from showing using javascript. They will now be able to pass an option into init called no_journeys. If set to true, we will not show a journeys banner for that page.

JIRA: https://branch.atlassian.net/browse/JOUR-720


aaustin:
Nice 👍

jsaleigh:
Nice! 👍

PR #415, opened 2017-01-19
INTENG-3144 remove css from previous banner

yntema:
A partner noticed that when clicking to a new page, a new banner would render, but some of the css was still on the page cause the content to get pushed down further than expected.

This code checks for the existence of #branch-banner-iframe style tag and removes it if it exists. Should not change any other functionality.


jsaleigh:
👍

rubinsingh:
👍  just tested and it works

PR #420, opened 2017-01-25
INTENG-3152 Make closeJourney smoother

yntema:
Removing branch-iframe-css outside of the animation setTimeout caused the banner to jump before disappearing. This PR places the css removal code within the setTimeout to prevent this.


rubinsingh:
👍

PR #424, opened 2017-02-07
fixes for compiler warnings

rubinsingh:
@yntema @scott-branchio this should help with the compiler warnings that we were seeing earlier.


yntema:
👍

PR #427, opened 2017-02-15
DANA-128 - add trackCommerceEvent() method

rubinsingh:
This PR adds the ability to send a purchase event from the WebSDK

Usage example:

```
var metadata =  { "foo": "bar" };

var commerce_data = {
    "revenue": 50.0,
    "currency": "USD",
    "transaction_id": "foo-transaction-id",
    "shipping": 0.0,
    "tax": 5.0,
    "affiliation": "foo",
    "products": [
        { "sku": "foo-sku-1", "name": "foo-item-1", "price": 45.00, "quantity": 1, "brand": "foo-brand",
            "category": "Electronics", 
            "variant": "foo-variant-1"
        }
        ,
        { "sku": "foo-sku-2", "price": 2.50, "quantity": 2}
    ],
};

branch.trackCommerceEvent('purchase', commerce_data, metadata, function(err){
    if(err){
        throw err;
    }
});
```

### DIFF
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -555,59 +555,47 @@ utils.getBrowserLanguageCode = function() {
 };
 
 /**
- * Returns the difference between two arrays. 
+ * Returns the difference between two arrays.
  */
 utils.calculateDiffBetweenArrays = function(original, toCheck) {
 	var diff = [];
 	toCheck.forEach(function(element) {
-		if(original.indexOf(element) == -1) {
+		if (original.indexOf(element) === -1) {
 			diff.push(element);
 		}
 	});
-
 	return diff;
-}
+};
 
-/** 
+/**
  * Validates the commerce-data object passed into branch.trackCommerceEvent().
  * If there are invalid keys present then it will report back what those keys are.
  * Note: The keys below are optional.
  */
 utils.validateCommerceData = function(commerceData) {
-	var allowedInRoot = ['common', 'type', 'transaction_id', 'currency', 'revenue', 'revenue_in_usd', 'exchange_rate', 'shipping', 'tax', 'coupon', 'affiliation', 'persona', 'products'];
-	var allowedInProducts = ['sku', 'name', 'price', 'quantity', 'brand', 'category', 'variant'];
+	var allowedInRoot = [ 'common', 'type', 'transaction_id', 'currency', 'revenue', 'revenue_in_usd', 'exchange_rate', 'shipping', 'tax', 'coupon', 'affiliation', 'persona', 'products' ];
+	var allowedInProducts = [ 'sku', 'name', 'price', 'quantity', 'brand', 'category', 'variant' ];
 
 	var invalidKeys = {
 		root: utils.calculateDiffBetweenArrays(allowedInRoot, Object.keys(commerceData)),
 		products: []
-	}
+	};
 
-	if(commerceData.hasOwnProperty('products')) {
+	if (commerceData.hasOwnProperty('products')) {
 		commerceData["products"].forEach(function(product) {
-			utils.merge(invalidKeys["products"], utils.calculateDiffBetweenArrays(allowedInProducts, Object.keys(product)));
+			invalidKeys['products'] = invalidKeys['products'].concat(utils.calculateDiffBetweenArrays(allowedInProducts, Object.keys(product)));
 		});
 	}
-
 	return invalidKeys;
-}
+};
 
-/** 
- * Builds an error message if partner provides invalid keys to commerce_data
- */
-utils.buildErrorMessageForCommerceData = function(invalidKeys) {
-	var errMessage = 'Please remove the following keys from ';
+utils.validCommerceEvents = [ 'purchase' ];
 
-	if (invalidKeys['root'].length) {
-		errMessage += 'the root of commerce_data: ' + invalidKeys['root'].join(', ');
-	}
+/** @type {Object<string,message>} */
+utils.commerceEventMessages = {
+	missingPurchaseEvent: 'event name is either missing, of the wrong type or not valid. Please specify \'purchase\' as the event name.',
+	missingCommerceData: 'commerce_data is either missing, of the wrong type or empty. Please ensure that commerce_data is constructed correctly.',
+	invalidKeysForRoot: 'Please remove the following keys from the root of commerce_data: ',
+	invalidKeysForProducts: 'Please remove the following keys from commerce_data.products: '
+};
 
-	if (invalidKeys['root'].length && invalidKeys['products'].length) {
-		errMessage += '. In addition, please remove the following keys from commerce_data.products: ' + invalidKeys['products'].join(', ') + '.';
-	}
-
-	else if (invalidKeys['products'].length) {
-		errMessage += 'commerce_data.products: ' + invalidKeys['products'].join(', ');
-	}
-
-	return errMessage;
-}
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -568,59 +568,59 @@ utils.calculateDiffBetweenArrays = function(original, toCheck) {
 	return diff;
 };
 
+var validCommerceEvents = [ 'purchase' ];
+
+var commerceEventMessages = {
+	missingPurchaseEvent: 'event name is either missing, of the wrong type or not valid. Please specify \'purchase\' as the event name.',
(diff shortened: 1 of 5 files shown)


src/1_utils.js
```diff
@@ -553,3 +553,61 @@ utils.getBrowserLanguageCode = function() {
...
+
+	if (invalidKeys['root'].length && invalidKeys['products'].length) {
+		errMessage += '. In addition, please remove the following keys from commerce_data.products: ' + invalidKeys['products'].join(', ') + '.';
+	}
+
+	else if (invalidKeys['products'].length) {
+		errMessage += 'commerce_data.products: ' + invalidKeys['products'].join(', ');
+	}
+
+	return errMessage;
```
derrickstaten: The ifs and else if seem a little odd here. Here's an alternative that is a bit more understandable to me, reading this for the first time:

```
if (invalidKeys['root'].length) {
	errMessage += 'the root of commerce_data: ' + invalidKeys['root'].join(', ');
	if (invalidKeys['products'].length) {
		'errMessage += '. In addition, please remove the following keys from ';
	}
}
if (invalidKeys['products'].length) {
	errMessage += 'commerce_data.products: ' + invalidKeys['products'].join(', ');
}
```

feel free to ignore this suggestion though

derrickstaten: Also, what's common with error messages is to not address every error in one pass. You could remove this helper function entirely and down in the original code just do 

```
if (invalidKeys["root"].length) {
	done('Please remove the following keys from the root of commerce_data: ' + invalidKeys['root'].join(', '));
}
else if (invalidKeys["products"].length) {
	done('Please remove the following keys from commerce_data.products: ' + invalidKeys['products'].join(', '));
}
```

You don't have to get it all in one pass, as nice as that is :) again, ultimately this is up to you and Gerrit, I'm definitely nitpicking.

derrickstaten:
I'm not as knowledgeable about the web sdk as Gerrit, so I'll let him review.

src/6_branch.js
```diff
@@ -1817,3 +1817,93 @@ Branch.prototype['autoAppIndex'] = wrap(callback_params.CALLBACK_ERR, function(d
...
+
+	var invalidKeys = utils.validateCommerceData(commerce_data);
+
+	if (invalidKeys["root"].length) {
+		return done(utils.commerceEventMessages.invalidKeysForRoot + invalidKeys['root'].join(', '));
+	}
+
+	else if (invalidKeys["products"].length) {
+		return done(utils.commerceEventMessages.invalidKeysForProducts + invalidKeys['products'].join(', '));
+	}
```
yntema: You should abstract all of this validation logic into a helper function that returns the appropriate message or throws the appropriate error

src/2_resources.js
```diff
@@ -297,3 +297,15 @@ resources.event = {
...
+
+resources.commerceEvent = {
+	destination: config.api_endpoint,
+	endpoint: "/v1/event",
+	method: utils.httpMethod.POST,
+	params: defaults({
+		"event": validator(true, validationTypes.STRING),
+		"metadata": validator(false, validationTypes.OBJECT),
+		"initial_referrer": validator(false, validationTypes.STRING),
+		"commerce_data": validator(true, validationTypes.OBJECT)
```
yntema: Are we handling this commerce_data param in the api yet? We need to make sure this is how the api wants it

src/1_utils.js
```diff
@@ -553,3 +553,49 @@ utils.getBrowserLanguageCode = function() {
 	}
 	return code;
 };
+
+/**
+ * Returns the difference between two arrays.
```
yntema: Be a little more descriptive here if possible

src/6_branch.js
yntema: Do we want to throw an error for any of these invalid situations?

src/6_branch.js
```diff
@@ -1817,3 +1817,93 @@ Branch.prototype['autoAppIndex'] = wrap(callback_params.CALLBACK_ERR, function(d
 		done(null);
 	}
 });
+
+/**
+ * @function Branch.trackCommerceEvent
+ * @param {String} event - _required_ - Name of the commerce event to be tracked. We currently 	support 'purchase' events.
```
yntema: fix spacing 'We currently         support'

src/1_utils.js
```diff
@@ -553,3 +553,49 @@ utils.getBrowserLanguageCode = function() {
...
+
+	var invalidKeys = {
+		root: utils.calculateDiffBetweenArrays(allowedInRoot, Object.keys(commerceData)),
+		products: []
+	};
+
+	if (commerceData.hasOwnProperty('products')) {
+		commerceData["products"].forEach(function(product) {
+			invalidKeys['products'] = invalidKeys['products'].concat(utils.calculateDiffBetweenArrays(allowedInProducts, Object.keys(product)));
+		});
```
yntema: I don't think this would work since commereData["products"] is an array of objects. You would be comparing the list from `allowedInProducts` which is an array of strings

src/2_resources.js
derrickstaten: yes

src/6_branch.js
derrickstaten: sorry @rubinsingh I told you to move some of it out of a separate function

src/1_utils.js
yntema: nvm, didn't see `Object.keys`

src/6_branch.js
rubinsingh: @yntema sounds good and I'll throw a hard error for all of these situations

src/1_utils.js
```diff
@@ -553,3 +553,74 @@ utils.getBrowserLanguageCode = function() {
...
+	if (!commerce_data || typeof commerce_data !== 'object' || Object.keys(commerce_data).length === 0) {
+		return utils.commerceEventMessages.missingCommerceData;
+	}
+
+	var invalidKeys = utils.validateCommerceDataKeys(commerce_data);
+
+	if (invalidKeys['root'].length) {
+		return utils.commerceEventMessages.invalidKeysForRoot + invalidKeys['root'].join(', ');
+	}
+
```
yntema: remove this extra line return

src/1_utils.js
```diff
@@ -553,3 +553,74 @@ utils.getBrowserLanguageCode = function() {
...
+utils.validCommerceEvents = [ 'purchase' ];
+
+/** @type {Object<string,message>} */
+utils.commerceEventMessages = {
+	missingPurchaseEvent: 'event name is either missing, of the wrong type or not valid. Please specify \'purchase\' as the event name.',
+	missingCommerceData: 'commerce_data is either missing, of the wrong type or empty. Please ensure that commerce_data is constructed correctly.',
+	invalidKeysForRoot: 'Please remove the following keys from the root of commerce_data: ',
+	invalidKeysForProducts: 'Please remove the following keys from commerce_data.products: '
+};
```
yntema: Neither of these need to be exported. You should declare them as variables and move them above `utils.validateCommerceEventParams`

src/1_utils.js
```diff
@@ -553,3 +553,74 @@ utils.getBrowserLanguageCode = function() {
...
+	});
+	return diff;
+};
+
+/**
+ * Validates the commerce-data object passed into branch.trackCommerceEvent().
+ * If there are invalid keys present then it will report back what those keys are.
+ * Note: The keys below are optional.
+ */
+utils.validateCommerceDataKeys = function(commerceData) {
```
yntema: doesn't need to be exported

src/1_utils.js
```diff
@@ -553,3 +553,74 @@ utils.getBrowserLanguageCode = function() {
...
+utils.validateCommerceEventParams = function(event, commerce_data) {
+	if (!event || typeof event !== 'string' || utils.validCommerceEvents.indexOf(event.toLowerCase()) === -1) {
+		return utils.commerceEventMessages.missingPurchaseEvent;
+	}
+
+	if (!commerce_data || typeof commerce_data !== 'object' || Object.keys(commerce_data).length === 0) {
+		return utils.commerceEventMessages.missingCommerceData;
+	}
+
+	var invalidKeys = utils.validateCommerceDataKeys(commerce_data);
```
yntema: I would rather see this just return the appropriate message string than an object with the the invalid properties. This seems confusing to me

yntema:
👍

PR #428, opened 2017-02-24
check whether session storage is native code, or added by a user

jsaleigh:


jsaleigh:
@yntema @scott-branchio note that this will need to be tested on all devices, if there is a decision to merge this

scott-branchio:
is there a JIRA for this?

PR #430, opened 2017-03-02
JOUR-768 - Use window.location instead of document.URL

yntema:
https://branch.atlassian.net/plugins/servlet/mobile#issue/JOUR-768


rubinsingh:
👍

PR #438, opened 2017-03-23
INTENG-3188 - v1/credits fix

rubinsingh:
This PR fixes .credits(). It ensures that the Branch Key and Identity is sent to v1/credits.


yntema:
👍

PR #439, opened 2017-03-28
Add data parameter to _publishEvent function

yntema:
This will give users the ability to receive the banner_id (branchViewId) in the listener callback when journeys events are published.


src/6_branch.js
```diff
@@ -210,11 +210,12 @@ Branch.prototype._referringLink = function() {
...
  */
-Branch.prototype._publishEvent = function(event) {
+Branch.prototype._publishEvent = function(event, data) {
 	for (var i = 0; i < this._listeners.length; i++) {
 		if (!this._listeners[i].event || this._listeners[i].event === event) {
-			this._listeners[i].listener(event);
+			this._listeners[i].listener(event, data);
 		}
 	}
 };
```
rubinsingh: Might want to add banner_id to https://github.com/BranchMetrics/web-branch-deep-linking/blob/master/src/6_branch.js#L1557 (the event that is published for didCallJourneyClose) as well

src/6_branch.js
```diff
@@ -1466,7 +1467,7 @@ Branch.prototype['addListener'] = function(event, listener) {
 
 /** =WEB
  * @function Branch.removeListener
- * @param {function(String)} listener - _required_ - Reference to the listening function you
+ * @param {function(String, Object)} listener - _required_ - Reference to the listening function you
```
rubinsingh: let's remove the additional Object param as the definition for removeListener has not changed

src/1_utils.js
```diff
@@ -38,7 +38,7 @@ utils.httpMethod = {
  * }} */
 utils.resource;
 
-/** @typedef {{listener: function(string):null, event: string}} */
+/** @typedef {{listener: function(string, Object):null, event: string}} */
```
rubinsingh: We could additionally describe the object as:

+/** @typedef {{listener: function(string, Object):null, event: string, data: object}} */

rubinsingh:
👍

PR #440, opened 2017-03-28
INFRA-1778 validate text input for branch.banner()

yntema:
A simple validation to make sure users aren't injecting html into the text fields


rubinsingh:
👍

PR #441, opened 2017-03-29
WEBSDK-40 - Exit whiteListJourneysLanguageData if data is null or undefined

yntema:
We were seeing errors on a partners site. When data is null, we were trying to execute Object.keys on it. This change exits early if data is null or undefined.


rubinsingh:
👍

PR #445, opened 2017-04-06
Support `link_identifier` in branch.init() options

tyfoo:
Adds the ability to send `link_identifier` as`branch.init` part of the `options` object:
```jsx
  branch.init(branch_key, { link_identifier: _branch_match_id }, callback);
```

The reasoning behind it is we currently use `branch.io` on a react/redux app and need to clear the query params before calling init, i.e.
```jsx
  const { router: { location: { query, pathname } = { } } = { } } = this.props;
  const link_identifier = query._branch_match_id;
  this.props.router.replace({ pathname }); // clears query string, and hence window.location.search
  // later in the code / component life-cycle
  window.branch.init(BRANCH_IO_KEY, { link_identifier }, (err = {}, data) => { ... });
```

### DIFF
--- a/example.html
+++ b/example.html
@@ -51,6 +51,10 @@
 				<h4>Session and Identity</h4>
 				<div class="group">
 					<button id="init" class="btn btn-success">.init()</button>
+					<button id="init-options" class="btn btn-success">.init(key, { link_identifier }, callback)</button>
+					<input class="example-input" type="text" id="link_identifier" class="form-control" placeholder="link_identifier">
+				</div>
+				<div class="group">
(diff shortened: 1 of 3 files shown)


rubinsingh:
@tyfoo thank you for submitting this PR. We have a similar feature coming out in the next release of the WebSDK so I'll go ahead and close this for now.

You'll be able to use it like so:

branch.init('<insert_branch_key>', { branch_match_id : '<insert_branch_match_id>' }, initCallback);

- https://github.com/BranchMetrics/web-branch-deep-linking/pull/462/files

PR #448, opened 2017-04-26
sending up screen width and height

rubinsingh:


src/6_branch.js
```diff
@@ -393,7 +393,9 @@ Branch.prototype['init'] = wrap(
 					"url": utils.getWindowLocation(),
 					"user_agent": navigator.userAgent,
 					"language": navigator.language,
-					"page_has_microdata": checkForMicroData()
+					"page_has_microdata": checkForMicroData(),
+					"screen_width": window.screen.width ? window.screen.width : 0,
+					"screen_height": window.screen.height ? window.screen.height : 0
```
yntema: - Are you sure we want to default to 0? I would confirm with whoever will be using this data
- You can do `"screen_width": screen.width || 0,`, don't need to include window

rubinsingh: @yntema we've decided to default to -1 instead

src/6_branch.js
```diff
@@ -393,7 +393,9 @@ Branch.prototype['init'] = wrap(
 					"url": utils.getWindowLocation(),
 					"user_agent": navigator.userAgent,
 					"language": navigator.language,
-					"page_has_microdata": checkForMicroData()
+					"page_has_microdata": checkForMicroData(),
+					"screen_width": screen.width ? screen.width : -1,
+					"screen_height": screen.height ? screen.height : -1
```
yntema: `"screen_width": screen.width || -1,`

yntema:
👍

PR #450, opened 2017-04-28
WEBSDK-43 - appends branch_key to ?_r request

rubinsingh:


src/6_branch.js
```diff
@@ -320,7 +320,7 @@ Branch.prototype['init'] = wrap(
 		var freshInstall = !sessionData || !sessionData['identity_id'];
 		self._branchViewEnabled = !!self._storage.get('branch_view_enabled');
 		var checkHasApp = function(sessionData, cb) {
-			var params_r = { "sdk": config.version };
+			var params_r = { "sdk": config.version, "branch_key": self.branch_key };
```
yntema: there is another place that `_r` is called, starting at line 453

rubinsingh: @yntema thanks for catching that. Updating now.

yntema:
👍

PR #455, opened 2017-05-19
JOUR-841

rubinsingh:


src/journeys_utils.js
```diff
@@ -36,6 +36,16 @@ journeys_utils.branch = null;
...
 journeys_utils.isJourneyDisplayed = false;
 
+journeys_utils.animationSpeed = 250;
+journeys_utils.animationDelay = 20;
+
+journeys_utils.exitAnimationDisabled = false;
+journeys_utils.entryAnimationDisabled = false;
+journeys_utils.journeyDismissed = false;
+journeys_utils.exitAnimationDisabledPreviously = false;
+journeys_utils.previousPosition = '';
```
yntema: could you add a quick comment on what these are for.

src/journeys_utils.js
```diff
@@ -258,7 +270,6 @@ journeys_utils.addIframeOuterCSS = function() {
 		var calculatedBodyMargin = +bannerMarginNumber + bodyMarginBottomNumber;
 		bodyMargin = 'margin-bottom: ' + calculatedBodyMargin + 'px';
 	}
-
 	// adds margin to the parent of div being inserted into
 	if (journeys_utils.divToInjectParents && journeys_utils.divToInjectParents.length > 0) {
```
yntema: We can remove the first check here since we are setting a default on line 132. It will have a value, so this is unnecesary.

yntema:
👍

PR #467, opened 2017-06-22
WEBSDK 59 - reset session data when link_identifier not available

rubinsingh:


src/6_branch.js
```diff
@@ -459,9 +459,10 @@ Branch.prototype['init'] = wrap(
 		};
 
 		if (sessionData && sessionData['session_id'] && !link_identifier) {
+			// resets data in session storage to prevent previous link click data from being returned to Branch.init()
+			session.update(self._storage, { "data": "" });
```
yntema: @rubinsingh do we want to set data to and empty string? or would it make more sense to set it to null?

yntema: or is data stringified json?

rubinsingh: @yntema thanks for noticing this. The reason why we want to make it an empty string is because most partners use the cdn version of the WebSDK which is always up-to-date with the latest code. In the past we have returned the previous session data which is a non-empty string. If we return a null, this could cause some sites throw errors.

yntema:
👍

constantx:
When will this be updated to npm?

PR #477, opened 2017-07-26
WEBSDK-74, WEBSDK-75, WEBSDK-76

yntema:
- don't call _r on safari 11 or newer
- send bfp from local as alternative in open call
- send branch_match_id if available


src/2_resources.js
```diff
@@ -93,7 +93,8 @@ resources.open = {
 	endpoint: "/v1/open",
 	method: utils.httpMethod.POST,
 	params: {
-		"browser_fingerprint_id": validator(true, branch_id),
+		"browser_fingerprint_id": validator(false, branch_id),
+		"alt_browser_fingerprint_id": validator(false, branch_id),
```
ahmednawar: can we change to `alternative`?

src/1_utils.js
```diff
@@ -661,3 +697,13 @@ utils.openGraphDataAsObject = function() {
...
 
+utils.appendBfpToLinkUrl = function(url, bfp) {
+	if (typeof url !== 'string') {
+		return null;
+	}
+	if (!bfp) {
+		return url;
+	}
+
+	return url + '?_t=' + bfp;
```
ahmednawar: can we check for the existence of a `?` before appending the bfp in case v1/url returns a query param? It doesn't happen right now but something similar happened on the mobile sdks before and it is still biting us.

src/2_resources.js
yntema: Good catch!

src/1_utils.js
yntema: sure can

src/2_resources.js
```diff
@@ -93,7 +93,8 @@ resources.open = {
 	endpoint: "/v1/open",
 	method: utils.httpMethod.POST,
 	params: {
-		"browser_fingerprint_id": validator(true, branch_id),
+		"browser_fingerprint_id": validator(false, branch_id),
+		"alternative_browser_fingerprint_id": validator(false, branch_id),
```
zeeshabnc: alternative_browser_fingerprint_id looks exactly the same as browser_fingerprint_id. Shouldn't it be the value from storage instead based on the ticket desc?

yntema: @zeeshabnc this file just defines the parameter type and validates the request. If you look at where open is called in `6_branch` you will see a difference and that `alternative_browser_fingerprint_id` is the value from storage

rubinsingh:
👍

rubinsingh:
👍

PR #482, opened 2017-08-02
passing page title and description to pageview event

rubinsingh:


src/6_branch.js
```diff
@@ -409,7 +409,9 @@ Branch.prototype['init'] = wrap(
 					"screen_width": screen.width || -1,
 					"screen_height": screen.height || -1,
 					"og_data": utils.openGraphDataAsObject(),
-					"hosted_deeplink_data": utils.scrapeHostedDeepLinkData()
+					"hosted_deeplink_data": utils.scrapeHostedDeepLinkData(),
+					"title": utils.scrapeTitle(),
+					"description": utils.scrapeDescription()
```
pccyang: scrapeXXX might be a good method name since websdk is open source I assume. Don't want to trigger an alarm to partners. getXXX() might be better.

pccyang: Can we also add canonical_url?

pccyang:
lgtm. I'll defer to @yntema to give the final thumb.

yntema:
@rubinsingh  Looks good. Fix conflicts and then we will wait until other deploy goes out.

zeeshabnc:
@rubinsingh  Henceforth can you please add the JIRA number into the commit message as the first line please. That way JIRA links them together and its easier to see the PR relation with JIRAs.

rubinsingh:
Thanks @zeeshabnc will do that going forward!

PR #484, opened 2017-08-07
WEBSDK-88 - Ensures that null values are not sent up for additional metadata keys

rubinsingh:
Ensures that null values are not sent up for additional metadata keys


pccyang:
Seems like trying to remove the null fields make the code a bit complicated. rubin and gerrit,  your thoughts on keeping the codes clean and return null fields or making it a bit complex to remove the null fields?
Or, rubin, can we make it a bit generic and have a utils method to remove null fields and empty object?

PR #492, opened 2017-08-14
[INFRA-2346] Deploy websdk

JaveriaK:
@yntema @rubinsingh @hubt 

This now has everything except for npm publish


JaveriaK:
npm publish also enabled, just not tested yet since we need to deploy for that

deployment/deploy.sh
```diff
@@ -93,30 +89,26 @@ else
...
-#git push origin $VERSION
+echo "Pushing changes back to repo ..."
+git push origin $CIRCLE_BRANCH
 
 # Send an update to slack channels
 DEPLOY_IMG=http://workshops.lewagon.org/assets/landing-2/deploy-button-5068ec2c575492ba428569111afe3ce6.jpg
 echo "Sending update to slack ..."
-#slackcli -t $SLACK_TOKEN -h int-eng -m "$CIRCLE_USERNAME Deploying WedSDK:$CIRCLE_BRANCH v$VERSION" -u websdk-deploy -i $DEPLOY_IMG
+#uncomment to send updates to int-eng
+#slackcli -t $SLACK_TOKEN -h int-eng -m "$CIRCLE_USERNAME Deployed WedSDK:$CIRCLE_BRANCH v$VERSION" -u websdk-deploy -i $DEPLOY_IMG
```
hubt: For slack notifications, we should do the same thing we do with other repos for consistency. When we look back at deployment histories, it'll be easier if we do it the same way. The script we use for other repos is in https://github.com/BranchMetrics/deploy/blob/master/scripts/rollbar.py
You can just copy that file in if you don't want to add the whole repo, but do remove the rollbar api keys from the repo and put them into environment variables because this is a public repo.

JaveriaK: added rollbar in addition, leaving slack cli for now, we can remove it later if rollbar is sufficient for updates.

hubt:
👍

hubt:
👍

PR #493, opened 2017-08-18
WEBSDK-67 - logEvent()

rubinsingh:


src/1_utils.js
```diff
@@ -733,3 +733,44 @@ utils.getAdditionalMetadata = function() {
...
 	return metadata && Object.keys(metadata).length > 0 ? metadata : null;
 };
+
+// v2/event utility functions
+
+var BRANCH_STANDARD_EVENTS = [ 'ADD_TO_CART', 'ADD_TO_WISHLIST', 'VIEW_CART', 'INITIATE_PURCHASE', 'ADD_PAYMENT_INFO', 'PURCHASE', 'SPEND_CREDITS', 'SEARCH', 'VIEW_CONTENT', 'VIEW_CONTENT_LIST', 'RATE', 'SHARE_CONTENT', 'COMPLETE_REGISTRATION', 'COMPLETE_TUTORIAL', 'ACHIEVE_LEVEL', 'UNLOCK_ACHIEVEMENTS' ];
+var BRANCH_STANDARD_EVENT_DATA = [ 'transaction_id', 'revenue', 'currency', 'shipping', 'tax', 'coupon', 'affiliation', 'search_query', 'description', 'product_categories' ];
+
+var isStandardEvent = function(eventName) {
+	return BRANCH_STANDARD_EVENTS.indexOf(eventName) > -1 ? true : false;
```
yntema: ` ? true : false;` is unnecessary. The previous statement will already return a boolean

src/1_utils.js
```diff
@@ -733,3 +733,44 @@ utils.getAdditionalMetadata = function() {
...
+};
+
+utils.determineV2Endpoint = function(eventName) {
+	return isStandardEvent(eventName) ? 'STANDARD' : 'CUSTOM';
+};
+
+utils.extractEventAndCustomData = function(eventAndCustomData) {
```
yntema: i might call this `separateEventAndCustomData`

src/1_utils.js
```diff
@@ -733,3 +733,44 @@ utils.getAdditionalMetadata = function() {
...
+
+	if (!eventAndCustomData || Object.keys(eventAndCustomData).length === 0) {
+		return null;
+	}
+	var customDataKeys = utils.calculateDiffBetweenArrays(BRANCH_STANDARD_EVENT_DATA, Object.keys(eventAndCustomData));
+	var customData = {};
+	var eventData = Object.assign({}, eventAndCustomData);
```
yntema: Object.assign is not supported on all browsers

src/1_utils.js
```diff
@@ -733,3 +733,44 @@ utils.getAdditionalMetadata = function() {
...
+		event_data: eventData
+	};
+};
+
+utils.validateParameterType = function(parameter, type) {
+	if (type === "array") {
+		return Array.isArray(parameter);
+	}
+	return typeof parameter === type && !Array.isArray(parameter);
+};
```
yntema: Could you add some tests for these utility functions?

src/3_api.js
```diff
@@ -290,7 +290,12 @@ Server.prototype.XHRRequest = function(url, data, method, storage, callback, nop
 			}
 			else if (req.status.toString().substring(0, 1) === '4' ||
 					req.status.toString().substring(0, 1) === '5') {
-				callback(new Error('Error in API: ' + req.status), null, req.status);
+				if (req.responseURL && req.responseURL.includes("v2/event")) {
+					callback(req.responseText, null, req.status);
+				}
```
yntema: Could you add a comment on why the `v2/event` error response is handled differently?

src/6_branch.js
```diff
@@ -1965,3 +1965,37 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
 	});
 	done();
 });
+/*
+ logEvent() documentation
+ */
```
yntema:
```
/**
 * logEvent() documentation
 */
```
Are you going add some documentation here?

src/6_branch.js
```diff
@@ -1965,3 +1965,37 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
...
+Branch.prototype['logEvent'] = wrap(callback_params.CALLBACK_ERR, function(done) {
+	var self = this;
```
yntema: I think the `self` declaration is unnecessary here

src/1_utils.js
```diff
@@ -733,3 +733,44 @@ utils.getAdditionalMetadata = function() {
...
+// v2/event utility functions
+
+var BRANCH_STANDARD_EVENTS = [ 'ADD_TO_CART', 'ADD_TO_WISHLIST', 'VIEW_CART', 'INITIATE_PURCHASE', 'ADD_PAYMENT_INFO', 'PURCHASE', 'SPEND_CREDITS', 'SEARCH', 'VIEW_CONTENT', 'VIEW_CONTENT_LIST', 'RATE', 'SHARE_CONTENT', 'COMPLETE_REGISTRATION', 'COMPLETE_TUTORIAL', 'ACHIEVE_LEVEL', 'UNLOCK_ACHIEVEMENTS' ];
+var BRANCH_STANDARD_EVENT_DATA = [ 'transaction_id', 'revenue', 'currency', 'shipping', 'tax', 'coupon', 'affiliation', 'search_query', 'description', 'product_categories' ];
+
+var isStandardEvent = function(eventName) {
+	return BRANCH_STANDARD_EVENTS.indexOf(eventName) > -1 ? true : false;
+};
+
+utils.determineV2Endpoint = function(eventName) {
```
yntema: i would call this `determineV2EventType` since it doesn't really return an endpoint

src/6_branch.js
```diff
@@ -1965,3 +1965,37 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
...
+	var name = arguments && utils.validateParameterType(arguments[1], 'string') ? arguments[1] : null;
```
yntema: We should probably throw an error if they don't include an event name right?

src/6_branch.js
```diff
@@ -1965,3 +1965,37 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
 	});
 	done();
 });
+/*
+ logEvent() documentation
+ */
+Branch.prototype['logEvent'] = wrap(callback_params.CALLBACK_ERR, function(done) {
```
yntema: Can't we just define all of the parameters now instead of using `arguments` to pull them out? Seems like something changed and now they are static

src/3_api.js
rubinsingh: @yntema our server returns helpful information when a partner sends up incorrect fields. That info appears in req.responseText.

yntema: Right, but could you add that in a code comment?

rubinsingh: 👍  ah thanks, I read that wrong

src/6_branch.js
rubinsingh: Yup, Derrick is working on some documentation and I'll add it here eventually.

src/1_utils.js
rubinsingh: 👍

src/1_utils.js
rubinsingh: 👍

src/6_branch.js
rubinsingh: 👍

src/1_utils.js
rubinsingh: 👍

src/1_utils.js
rubinsingh: 👍

src/6_branch.js
rubinsingh: @yntema our request validation rule will take care of this. 

"name": validator(true, validationTypes.STRING)

src/1_utils.js
```diff
@@ -733,3 +733,43 @@ utils.getAdditionalMetadata = function() {
...
+		delete eventAndCustomData[key];
+	}
+
+	return {
+		custom_data: customData,
+		event_data: eventAndCustomData
+	};
+};
+
+utils.validateParameterType = function(parameter, type) {
```
yntema: I would first check that the parameter exists incase you are looking to validate an object and null get's passed in. `typeof null === 'object'` turns out to be true

src/6_branch.js
```diff
@@ -1965,3 +1965,36 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
 	});
 	done();
 });
+/*
+ logEvent() documentation
+ */
+Branch.prototype['logEvent'] = wrap(callback_params.CALLBACK_ERR, function(done, name, eventAndCustomData, contentItems) {
+	name = name && utils.validateParameterType(name, 'string') ? name : null;
```
yntema: you should just check if the parameter exists within the `validateParameterType` function. then you can remove the additional checks

src/6_branch.js
```diff
@@ -1965,3 +1965,36 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
...
+	var extractedEventAndCustomData = utils.separateEventAndCustomData(eventAndCustomData);
+
+	var eventCallback = function(err, data) {
+		return done(err || null);
+	};
+
+	if (endpoint === 'STANDARD') {
+		this._api(resources.logStandardEvent, {
+			"name": name,
+			"custom_data": safejson.stringify(extractedEventAndCustomData && extractedEventAndCustomData['custom_data'] || {}),
```
yntema: Can you double check that we want to pass an empty object as opposed to null?

src/6_branch.js
```diff
@@ -1965,3 +1965,36 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
...
+		}, eventCallback);
+	}
+	else {
+		this._api(resources.logCustomEvent, {
+			"name": name,
+			"custom_data": safejson.stringify(extractedEventAndCustomData && extractedEventAndCustomData["custom_data"] || {}),
+			"event_data": safejson.stringify(extractedEventAndCustomData && extractedEventAndCustomData['event_data'] || {}),
+			"content_items": safejson.stringify(contentItems || [])
+		}, eventCallback);
+	}
```
yntema: Looks like both of these api calls require the exact same object to be passed in. Could you abstract that piece out to remove duplicating the same code?

src/1_utils.js
rubinsingh: 👍

src/6_branch.js
rubinsingh: 👍

src/6_branch.js
rubinsingh: Just the endpoints are different so I've created some logic to determine which endpoint to route the request to. I was doing what I just noted earlier but saw some warnings when running make. I can't reproduce those warnings so I reverted to the previous approach.

src/6_branch.js
rubinsingh: Spoke with the team about this and they would like an empty object sent up. Name is required so if it's null our validation rule will take over and throw an error.

src/6_branch.js
```diff
@@ -786,6 +786,132 @@ Branch.prototype['track'] = wrap(callback_params.CALLBACK_ERR, function(done, ev
...
+ *    content_items,
+ *    function(err) { console.log(err); }
+ *);
+ * ```
+ * ___
+ */
+/*** +TOC_ITEM #logevent-callback &.logEvent()& ^ALL ***/
+Branch.prototype['logEvent'] = wrap(callback_params.CALLBACK_ERR, function(done, name, eventAndCustomData, contentItems) {
+	name = utils.validateParameterType(name, 'string') ? name : null;
+	eventAndCustomData = utils.validateParameterType(eventAndCustomData, 'object') ? eventAndCustomData : null;
```
yntema: @rubinsingh what are your thoughts on just calling this `eventData`? Everything passed here is event data. I feel like eventAndCustomData is a confusing name

dmitrig01: we probably want to split out eventData, which is validated against a fixed schema, from customData, which lands in a separate place and is validated against a different schema

src/1_utils.js
```diff
@@ -750,6 +750,62 @@ utils.removePropertiesFromObject = function(objectToModify, keysToRemove) {
...
+// Requires a reference to the branch object to access information such as browser_fingerprint_id
+utils.getUserData = function(branch) {
+	var user_data = {};
+	user_data = utils.addPropertyIfNotNull(user_data, "http_origin", document.URL);
+	user_data = utils.addPropertyIfNotNull(user_data, "user_agent", navigator.userAgent);
+	user_data = utils.addPropertyIfNotNull(user_data, "language", navigator.language);
+	user_data = utils.addPropertyIfNotNull(user_data, "screen_width", screen.width);
+	user_data = utils.addPropertyIfNotNull(user_data, "screen_height", screen.height);
+	user_data = utils.addPropertyIfNotNull(user_data, "http_referrer", document.referrer);
+	user_data = utils.addPropertyIfNotNull(user_data, "identity_id", branch.identity_id);
```
dmitrig01: i don't believe we're collecting identity id, we do collect developer_identity though.

src/2_resources.js
```diff
@@ -310,3 +310,32 @@ resources.commerceEvent = {
...
+
+resources.logCustomEvent = {
+	destination: config.api_endpoint,
+	endpoint: "/v2/event/custom",
+	method: utils.httpMethod.POST,
+	params: {
+		"name": validator(true, validationTypes.STRING),
+		"user_data": validator(true, validationTypes.STRING),
+		"custom_data": validator(false, validationTypes.STRING),
+		"event_data": validator(false, validationTypes.STRING),
```
dmitrig01: custom event will not have event_data or content_items

src/6_branch.js
```diff
@@ -798,6 +798,132 @@ Branch.prototype['track'] = wrap(callback_params.CALLBACK_ERR, function(done, ev
...
+ *    "$address_street": "230 South LaSalle Street",
+ *    "$address_city": "Chicago",
+ *    "$address_region": "IL",
+ *    "$address_country": "US",
+ *    "$address_postal_code": "60604",
+ *    "$latitude": 12.07,
+ *    "$longitude": -97.5,
+ *    "$image_captions": [ "my_img_caption1", "my_img_caption_2" ],
+ *    "$condition": "NEW",
+ *    "$custom_fields": "{\"foo1\":\"bar1\",\"foo2\":\"bar2\"}"
```
dmitrig01: custom field are just listed inline, e.g.:
```
{
  "$quantity":1,
  "foo1":"bar1"
}
```

src/1_utils.js
```diff
@@ -750,6 +750,62 @@ utils.removePropertiesFromObject = function(objectToModify, keysToRemove) {
...
+	user_data = utils.addPropertyIfNotNull(user_data, "browser_fingerprint_id", branch.browser_fingerprint_id);
+	user_data = utils.addPropertyIfNotNull(user_data, "developer_identity", branch.identity);
+	user_data = utils.addPropertyIfNotNull(user_data, "sdk", branch.sdk);
```
dmitrig01: let's split this into two - can you send "web" as SDK, and the version as sdk_version?

rubinsingh:
@dmitrig01 @yntema would you mind giving this another pass?

src/1_utils.js
```diff
@@ -586,15 +586,21 @@ utils.getHostedDeepLinkData = function() {
...
+			code = navigator.languages[0];
+		}
+		else if (navigator.language) {
+			code = navigator.language;
+		}
+		code = code.substring(0, 2).toUpperCase();
 	}
 	catch (e) {
-		code = 'en';
+		code = null;
```
zeeshabnc: Why are we changing the default language in case of exception?

src/1_utils.js
```diff
@@ -750,6 +756,61 @@ utils.removePropertiesFromObject = function(objectToModify, keysToRemove) {
 	}
 };
 
+// v2/event utility functions
+
+var BRANCH_STANDARD_EVENTS = [ 'ADD_TO_CART', 'ADD_TO_WISHLIST', 'VIEW_CART', 'INITIATE_PURCHASE', 'ADD_PAYMENT_INFO', 'PURCHASE', 'SPEND_CREDITS', 'SEARCH', 'VIEW_ITEM', 'VIEW_ITEMS', 'RATE', 'SHARE', 'COMPLETE_REGISTRATION', 'COMPLETE_TUTORIAL', 'ACHIEVE_LEVEL', 'UNLOCK_ACHIEVEMENT' ];
+var BRANCH_STANDARD_EVENT_DATA = [ 'transaction_id', 'revenue', 'currency', 'shipping', 'tax', 'coupon', 'affiliation', 'search_query', 'description' ];
```
zeeshabnc: @dmitrig01  Is this intended to have all the fields listed under event_data in the EO spreadsheet? I see a subset listed here, how do we know that we have the right subset. just trying to get up to speed here.

src/1_utils.js
rubinsingh: @dmitrig01 it turns out that this function is used by Journeys as well so we'd rather send back 'en' to ensure that Journeys render something readable in English in the case of an exception. 

I'd like to revert this back to 'en' if possible.

dmitrig01: can we leave that in the journeys-specific section? I would rather send no language than the wrong language in v2/event.

rubinsingh: Yup, we can do that!

zeeshabnc:
Looks good to me. Since Gerrit already has given a thumb and dmitri has reviewed too, here is the final 👍 Lets not merge this in util it passes QA.

PR #495, opened 2017-08-23
provides Journey data to event listeners

rubinsingh:


src/branch_view.js
```diff
@@ -23,12 +23,12 @@ function checkPreviousBanner() {
  * @param {boolean} hasApp
  */
 function renderHtmlBlob(parent, html, hasApp) {
-	journeys_utils.branch._publishEvent('willShowJourney', { 'banner_id': journeys_utils.branchViewId });
+	journeys_utils.branch._publishEvent('willShowJourney', journeys_utils.journeyLinkData);
```
jsaleigh: This line is publishing an event, before the journeyLinkData is set in line 31

jsaleigh:
@rubinsingh My only concern is if events are published before the link data is set. If you can show that this won't happen, this looks good to me.

src/branch_view.js
```diff
@@ -23,12 +23,12 @@ function checkPreviousBanner() {
...
 
 	var ctaText = hasApp ? 'OPEN' : 'GET';
 
 	journeys_utils.setPositionAndHeight(html);
-
+	journeys_utils.setJourneyLinkData(html);
```
jsaleigh: If you move line 26 under this, it looks like this will work.  I would love unit tests for this :)

jsaleigh:
lgtm 👍

PR #510, opened 2017-09-13
JOUR-911 - Allows users to target Journeys based on hosted deep link data

rubinsingh:


pccyang:
lgtm 👍 
you might want to have another reviewer who is familiar websdk

src/6_branch.js
```diff
@@ -767,6 +767,11 @@ Branch.prototype['track'] = wrap(callback_params.CALLBACK_ERR, function(done, ev
 
 	options = options || {};
 
+	if (event && event === "pageview") {
```
yntema: you don't need to check event:
`if (event === "pageview") {`

PR #511, opened 2017-09-15
Revert "JOUR-911 - Allows users to target Journeys based on hosted deep link data"

rubinsingh:
Reverts BranchMetrics/web-branch-deep-linking#510


yntema:
👍

PR #519, opened 2017-09-28
[WEBSDK-98] Adding no-cache cache-control headers

JaveriaK:
@hubt @rubinsingh


hubt:
👍 
Note that you're setting cache-control on the s3 files, but the CDN may override this behavior. So be sure to test the caching behavior through the CDN and not just the s3 bucket.

raravena80:
I believe you can do invalidation lists in Cloudfront, however as far as I know this precedes after the TTL has expired.

JaveriaK:
The CDN for this is set to use origin cache headers. We tested with the cdn link for performance and it seems to work fine.

PR #521, opened 2017-10-04
fixes bugs in cookie storage method

rubinsingh:


src/2_storage.js
```diff
@@ -162,13 +162,13 @@ var cookies = function(perm) {
 			return returnCookieObject;
 		},
 		get: function(key) {
```
zeeshabnc: Can we add a unit test on this, seems like a relatively simple unit test should build our confidence here and prevent regressions.

rubinsingh: 👍  I've added unit tests for all of the cookie storage methods and have fixed a few bugs in the process.

src/2_storage.js
```diff
@@ -136,39 +136,41 @@ BranchStorage.prototype['session'] = function() {
...
 	};
 	return {
 		getAll: function() {
-			var cookieArray = document.cookie.split(';');
 			var returnCookieObject = { };
+			var cookieArray = document.cookie.split(';');
 			for (var i = 0; i < cookieArray.length; i++) {
-				var cookie = cookieArray[i].replace(' ', '');
-				cookie = cookie.substring(0, cookie.length);
-				if (cookie.indexOf(BRANCH_KEY_PREFIX) !== -1) {
```
yntema: any reason we got rid of this `BRANCH_KEY_PREFIX` check?

src/2_storage.js
```diff
@@ -136,39 +136,41 @@ BranchStorage.prototype['session'] = function() {
...
 		},
 		get: function(key) {
-			var keyEQ = prefix(key) + '=';
+			key = prefix(key);
 			var cookieArray = document.cookie.split(';');
 			for (var i = 0; i < cookieArray.length; i++) {
-				var cookie = cookieArray[i];
-				cookie = cookie.substring(1, cookie.length);
-				if (cookie.indexOf(keyEQ) === 0) {
-					return retrieveValue(cookie.substring(keyEQ.length, cookie.length));
```
yntema: the `retrieveValue` function will coerce a boolean. Are you sure we no longer need that? If we do a check for false and the value is "false", it may cause some unexpected behavior

src/2_storage.js
```diff
@@ -177,25 +179,15 @@ var cookies = function(perm) {
...
-						deleteCookie(cookie);
-					}
-					else if (perm && cookie.indexOf('branch_expiration_date=') > 0) {
-						deleteCookie(cookie);
-					}
-				}
+				var cookie = cookieArray[i].trim();
+				var firstEqualSign = cookie.indexOf("=");
+				var cookieName = cookie.substring(0, firstEqualSign);
+				removeCookie(cookieName, true);
```
yntema: seems like this would remove all cookies on the sites domain and not just the ones branch adds. I think we need to add back the additional checks. 

Also, is there a reason we are no longer using expiration dates on these?

src/2_storage.js
rubinsingh: I've added this check back in:

https://github.com/BranchMetrics/web-branch-deep-linking/pull/521/files#diff-3351e3fcc07fe61dec4286479ff4455dR70

Anytime we want to send the value of a cookie back, the retrieveValue function will ensure that boolean values are properly returned.

src/2_storage.js
rubinsingh: I've included this check over at:

https://github.com/BranchMetrics/web-branch-deep-linking/blob/9fb097ae05626a8bfe79668d2f3fdc66de54def2/src/2_storage.js#L61

There is now a separate function which detects whether a cookie is a Branch cookie. We use it at the following locations:

https://github.com/BranchMetrics/web-branch-deep-linking/blob/9fb097ae05626a8bfe79668d2f3fdc66de54def2/src/2_storage.js#L173

https://github.com/BranchMetrics/web-branch-deep-linking/blob/9fb097ae05626a8bfe79668d2f3fdc66de54def2/src/2_storage.js#L200

src/2_storage.js
rubinsingh: There is now a check to ensure that we don't remove all cookies on the site's domain:

https://github.com/BranchMetrics/web-branch-deep-linking/blob/9fb097ae05626a8bfe79668d2f3fdc66de54def2/src/2_storage.js#L200

There are test cases for this as well:

https://github.com/BranchMetrics/web-branch-deep-linking/blob/9fb097ae05626a8bfe79668d2f3fdc66de54def2/test/2_storage.js#L99-L101

https://github.com/BranchMetrics/web-branch-deep-linking/blob/9fb097ae05626a8bfe79668d2f3fdc66de54def2/test/2_storage.js#L180-L199

rubinsingh: Also, the expiration date area below is never hit since we don't use 'permcookie' as a storage option in the SDK:

https://github.com/BranchMetrics/web-branch-deep-linking/blob/master/src/2_storage.js#L142-L147

This means that we don't have to check for anything related to 'branch_expiration_date'

yntema:
👍

PR #523, opened 2017-10-10
JOUR-978 - Adds support for Twitter meta tag scraping

rubinsingh:


src/1_utils.js
```diff
@@ -546,37 +546,86 @@ utils.getOpenGraphContent = function(property, content) {
 
 /**
  * Search for hosted deep link data on the page, as outlined here https://dev.branch.io/getting-started/hosted-deep-link-data/guide/#adding-metatags-to-your-site
- * Also searches for applink tags, i.e. <meta property="al:ios:url" content="applinks://docs" />
+ * Also searches for twitter and applinks tags, i.e. <meta property="al:ios:url" content="applinks://docs" />, <meta name="twitter:app:url:googleplay" content="twitter://docs">
  */
 utils.getHostedDeepLinkData = function() {
```
yntema: What do you think about breaking this into at least two functions to make it easier to test. First one can be a function that just gets the list of meta tags on the page. The second accepts a list of meta tags and then handles the rest of the functionality. That way you can test a bunch of different scenarios instead of just one

yntema: Or maybe just make this function accept a list of metatags and you can just pass them in after calling `document.getElementsByTagName('meta')`.

rubinsingh: 👍  I'll break this code up so that we can test it better.

yntema:
👍

PR #524, opened 2017-10-12
WEBSDK-89 - Future proofing utils.clickIdFromLink()

rubinsingh:


src/1_utils.js
```diff
@@ -247,7 +247,13 @@ utils.cleanLinkData = function(linkData) {
  * @param {String} link
  */
 utils.clickIdFromLink = function(link) {
-	return link ? link.substring(link.lastIndexOf('/') + 1, link.length) : null;
+	var elem = document.createElement("a");
+	elem.href = link;
```
yntema: if link is not a valid url, the href and pathname will use the current page url info which might cause issues. Is the reason you are using an a tag so you can return the search string?

rubinsingh: Thanks for the heads up on this! According to our backend engineers, the parameter will either contain a regular URL or a pathname + query params. This means that the code will continue to work if we pass in something like "/clickid?test_key=test_val". For other cases we are now going to return immediately (ie when link is undefined, null, an empty string or of the wrong type).

yntema:
👍

PR #535, opened 2017-12-04
JOUR-1012 - allows users to target Journeys users via metadata

rubinsingh:


src/1_utils.js
```diff
@@ -934,3 +934,15 @@ utils.convertObjectValuesToString = function(objectToConvert) {
 	return objectToConvert;
 };
 
+// Adds user-supplied metadata object from init to additionalMetadata['hosted_deeplink_data'] for Journeys targeting
+utils.addMetadataFromInitToHostedMetadata = function(metadataFromInit, additionalMetadata) {
```
yntema: I would call this `mergeHostedDeeplinkData`. merge makes sense since that's what it's actually doing.  hosted_deeplink_data is what is actually getting merged in. This function is also used in `track` so using init is not accurate

src/6_branch.js
```diff
@@ -398,6 +398,12 @@ Branch.prototype['init'] = wrap(
 			}
 
 			var additionalMetadata = utils.getAdditionalMetadata();
```
yntema: You should update getAdditionalMetadata to return an empty object instead of null if there is no hosted deeplink data so you don't have to do the extra check on line 403

src/1_utils.js
yntema: I would also have `additionalMetadata` as the first argument to be more consistent with out merge function which has the `to` object first

yntema: Could you also return the results? I would prefer to make this a pure function with no side effects

src/6_branch.js
```diff
@@ -398,6 +398,12 @@ Branch.prototype['init'] = wrap(
...
+			var metadataFromInit = options && options['metadata'] && utils.validateParameterType(options['metadata'], "object") && Object.keys(options['metadata']).length > 0 ? options['metadata'] : null;
```
yntema: options has already been defaulted to an empty object on line 302 so the options check is unnecessary. you can also remove options['metadata'] because that check will be handled in utils.validateParameterType

yntema:
👍

PR #536, opened 2017-12-06
WEBSDK-116 - fixes event listener bug

rubinsingh:
The following scenario:

```
var listener = function(event, data) { console.log(event, data); }
branch.addListener(listener);
```

was not working correctly before

### DIFF
--- a/test/6_branch.js
+++ b/test/6_branch.js
@@ -1433,7 +1433,7 @@ describe('Branch', function() {
 				1,
 				'one listener listening with no event specified'
 			);
-			assert.strictEqual(listenerFired, 1, 'observer fired once');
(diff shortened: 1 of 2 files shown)


test/6_branch.js
```diff
@@ -1433,7 +1433,7 @@ describe('Branch', function() {
 				1,
 				'one listener listening with no event specified'
 			);
-			assert.strictEqual(listenerFired, 1, 'observer fired once');
+			assert.strictEqual(listenerFired, 2, 'historically, observer fired twice');
```
yntema: Should we be firing a listener when no event is specified?

yntema:
👍

PR #544, opened 2018-02-05
Update README.md to use`BRANCH_KEY` as a placeholder instead of `BRANCH KEY`

jakecadams:
using `BRANCH_KEY` as a placeholder is helpful for selecting and replacing, given it will be selected as one item instead of two (`BRANCH` + `KEY`)

@rubinsingh @yntema


yntema:
@jakecadams , the README.md is autogenerated, so this change needs to be made in https://github.com/BranchMetrics/web-branch-deep-linking/blob/master/docs/1_intro.md

PR #545, opened 2018-02-06
Fixed typo in README.md

javiercr:


yntema:
@javiercr nice catch! Our README.md is autogenerated, so this would fix should happen here: https://github.com/BranchMetrics/web-branch-deep-linking/blob/master/src/6_branch.js#L1696

PR #556, opened 2018-04-25
JOUR-1806 - willShowJourney Update

TheMarshal99:
Updated renderHtmlBlob to fire 'willShowJourney' event after the Journey iframe has been built/added to the page.

JOUR-1806


TheMarshal99:
Ticketmaster is using this change already and it addresses their needs. It changes the purpose of the method from "A Journey is about to be shown" to "A Journey is on the page and about to animate in", thus allowing them to modify the position of the Journey before its animation has completed.

PR #557, opened 2018-05-05
DEVEX-325 & DEVEX-388 - GDPR functionality

rubinsingh:
Does not include changes to branch.link()

### DIFF
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -23,7 +23,9 @@ utils.timeout = 5000; // Duration in milliseconds that the system should wait fo
 utils.gdpr = {
 	tracking_disabled: false,
 	white_listed_endpoints_with_data: {
-		'/v1/open': { 'link_identifier':'\\d+' }
+		'/v1/open': { 'link_identifier':'\\d+' },
+		'/v1/event': { 'event': 'pageview' },
+		'/v1/branchview': {}
 	},
 	allow_errors_in_callback: false
 };
@@ -81,10 +83,7 @@ utils.cleanApplicationSessionAndCookieStorage = function(branch) {
 		// Sets an empty object for branch_session and branch_session_first in local/sessionStorage
 		session.set(self._storage, data, true);
 	}
-
-	// Attempts to delete _s cookie
-	var expires = 'Thu, 01 Jan 1970 00:00:01 GMT';
-	document.cookie = '_s' + '=; expires=' + expires + '; path=/;';
+	// a user will need to explicitly opt out from _s cookie
 };
 
 /** @typedef {{data:?string, referring_identity:?string, identity:?string, has_app:?boolean}} */
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -20,18 +20,18 @@ utils.retries = 2; // Value specifying the number of times that a Branch API cal
 utils.retry_delay = 200; // Amount of time in milliseconds to wait before re-attempting a timed-out request to the Branch API.
 utils.timeout = 5000; // Duration in milliseconds that the system should wait for a response before considering any Branch API call to have timed out.
 
-utils.gdpr = {
-	tracking_disabled: false,
-	white_listed_endpoints_with_data: {
+utils.userPreferences = {
+	trackingDisabled: false,
+	whiteListedEndpointsWithData: {
 		'/v1/open': { 'link_identifier':'\\d+' },
 		'/v1/event': { 'event': 'pageview' },
 		'/v1/branchview': {}
 	},
-	allow_errors_in_callback: false
+	allowErrorsInCallback: false
 };
 
 // Used by 3_api.js to determine whether a request should be blocked
-utils.gdpr.shouldBlockRequestInGDPRMode = function(url, requestData) {
+utils.userPreferences.shouldBlockRequest = function(url, requestData) {
 
 	var urlParser = document.createElement('a');
 	urlParser.href = url;
@@ -43,7 +43,7 @@ utils.gdpr.shouldBlockRequestInGDPRMode = function(url, requestData) {
 		urlPath = '/' + urlPath;
 	}
 
-	var whiteListedEndpointWithData = utils.gdpr.white_listed_endpoints_with_data[urlPath];
+	var whiteListedEndpointWithData = utils.userPreferences.whiteListedEndpointsWithData[urlPath];
 
 	if (!whiteListedEndpointWithData) {
 		return true;
@@ -66,7 +66,7 @@ utils.gdpr.shouldBlockRequestInGDPRMode = function(url, requestData) {
 };
 
 // Removes PII when a user disables tracking
-utils.cleanApplicationSessionAndCookieStorage = function(branch) {
+utils.cleanApplicationAndSessionStorage = function(branch) {
 	if (branch) {
 		// clears PII from global Branch object
 		branch.device_fingerprint_id = null;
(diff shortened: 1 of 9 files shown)


src/6_branch.js
```diff
@@ -2024,4 +2051,41 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
...
+ *
+ * Allows User to Remain Private
+ *
+ * This will prevent any Branch requests from being sent across the network, except for the case of deep linking.
+ * If someone clicks a Branch link, but has expressed not to be tracked, we will return deep linking data back to the
+ * client but without tracking information.
+ *
+ * In do-not-track mode, you will still be able to create links and display Journeys however, they will not have identifiable
+ * information associated to them. You can change this behavior at any time, by calling the aforementioned function.
+ * This information will be saved and persisted.
```
E-B-Smith: "The do-not-track mode state is persistent: it is saved for the user across browser sessions for the web site."

E-B-Smith:
I added a single documentation suggestion.  Otherwise it looks like it's in the spirit of the iOS/Android versions.

src/1_utils.js
```diff
@@ -20,6 +20,72 @@ utils.retries = 2; // Value specifying the number of times that a Branch API cal
...
+};
+
+// Used by 3_api.js to determine whether a request should be blocked
+utils.gdpr.shouldBlockRequestInGDPRMode = function(url, requestData) {
+
+	var urlParser = document.createElement('a');
+	urlParser.href = url;
+	var urlPath = urlParser.pathname;
+
+	if (urlPath && utils.gdpr.white_listed_endpoints_with_data.hasOwnProperty(urlPath) === false
```
yntema: You can simplify this by just doing:
`if (!utils.gdpr.white_listed_endpoints_with_data[urlPath]) {`

I would just pull `utils.gdpr.white_listed_endpoints_with_data[urlPath]` into a variable since you use it multiple times

src/1_utils.js
```diff
@@ -20,6 +20,72 @@ utils.retries = 2; // Value specifying the number of times that a Branch API cal
...
+	) {
+		return true;
+	}
+	else if (
+		urlPath &&
+		utils.gdpr.white_listed_endpoints_with_data.hasOwnProperty(urlPath) &&
+		Object.keys(utils.gdpr.white_listed_endpoints_with_data[urlPath]).length > 0) {
```
yntema: No need to check if urlPath exists. I would use the variable you pull out from above to make this a little more concise

src/1_utils.js
```diff
@@ -20,6 +20,72 @@ utils.retries = 2; // Value specifying the number of times that a Branch API cal
...
+		}
+	}
+	return false;
+};
+
+// Removes PII when a user disables tracking
+utils.cleanApplicationSessionAndCookieStorage = function(branch) {
+	if (branch) {
+		// clears PII from global Branch object
+		var self = branch;
```
yntema: this variable declaration doesn't seem necessary

src/6_branch.js
```diff
@@ -255,7 +259,7 @@ Branch.prototype._publishEvent = function(event, data) {
  * | retry_delay | *optional* - `integer `. Amount of time in milliseconds to wait before re-attempting a timed-out request to the Branch API. Default 200 ms.
  * | timeout | *optional* - `integer`. Duration in milliseconds that the system should wait for a response before considering any Branch API call to have timed out. Default 5000 ms.
  * | metadata | *optional* - `object`. Key-value pairs used to target Journeys users via the "is viewing a page with metadata key" filter.
- *
+ * | disable_tracking | *optional* - `boolean`. true or false depending on whether the user would like to remain private or not.
```
yntema: This should be `tracking_disabled`

src/6_branch.js
```diff
@@ -2024,4 +2051,41 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
...
+ * If someone clicks a Branch link, but has expressed not to be tracked, we will return deep linking data back to the
+ * client but without tracking information.
+ *
+ * In do-not-track mode, you will still be able to create links and display Journeys however, they will not have identifiable
+ * information associated to them. You can change this behavior at any time, by calling the aforementioned function.
+ * This information will be saved and persisted.
+ * ___
+ */
+/*** +TOC_HEADING &User Privacy& ^WEB ***/
+/*** +TOC_ITEM #disableTrackingdisableTracking &.disableTracking()& ^WEB ***/
```
yntema: `#disableTrackingdisableTracking` -> `#disableTracking`

src/6_branch.js
```diff
@@ -2024,4 +2051,41 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
 	done();
 });
 
-
+/**
+ * @function Branch.disableTracking
+ * @param {Boolean} disableTracking - _required_ - true or false depending on whether the user would like to remain private or not
```
yntema: We should explicitly state that passing in true will disable tracking and passing false will enable tracking and that it will overwrite the options passed into init

src/6_branch.js
```diff
@@ -2024,4 +2051,41 @@ Branch.prototype['trackCommerceEvent'] = wrap(callback_params.CALLBACK_ERR, func
...
+ *
+ * In do-not-track mode, you will still be able to create links and display Journeys however, they will not have identifiable
+ * information associated to them. You can change this behavior at any time, by calling the aforementioned function.
+ * This information will be saved and persisted.
+ * ___
+ */
+/*** +TOC_HEADING &User Privacy& ^WEB ***/
+/*** +TOC_ITEM #disableTrackingdisableTracking &.disableTracking()& ^WEB ***/
+Branch.prototype['disableTracking'] = wrap(callback_params.CALLBACK_ERR, function(done, disableTracking) {
+	var self = this;
```
yntema: `var self = this;` doesn't seem necessary

src/6_branch.js
yntema: Also looks like below they have they option to pass in no arguments and it would be the same as passing in true. We should spell that out here if that's the case

src/1_utils.js
```diff
@@ -20,6 +20,76 @@ utils.retries = 2; // Value specifying the number of times that a Branch API cal
...
+		urlPath = '/' + urlPath;
+	}
+
+	var whiteListedEndpointWithData = utils.userPreferences.whiteListedEndpointsWithData[urlPath];
+
+	if (!whiteListedEndpointWithData) {
+		return true;
+	}
+	else if (
+		whiteListedEndpointWithData &&
```
yntema: You can remove this `whiteListedEndpointWithData` check

src/3_api.js
```diff
@@ -326,6 +326,22 @@ Server.prototype.XHRRequest = function(url, data, method, storage, callback, nop
...
+		if (data.hasOwnProperty("alternative_browser_fingerprint_id")) {
+			delete data['alternative_browser_fingerprint_id'];
+		}
+		if (data.hasOwnProperty("identity_id")) {
+			delete data['identity_id'];
+		}
+		if (data.hasOwnProperty("session_id")) {
+			delete data['session_id'];
+		}
+	}
```
yntema: Suggestion:
```
if (utils.userPreferences.trackingDisabled) {
	var PII = ['browser_fingerprint_id','alternative_browser_fingerprint_id','identity_id','session_id']
	for (var i in PII) {
		delete data[PII[i]];
	}
}
```

src/1_utils.js
```diff
@@ -20,6 +20,76 @@ utils.retries = 2; // Value specifying the number of times that a Branch API cal
...
+	whiteListedEndpointsWithData: {
+		'/v1/open': { 'link_identifier':'\\d+' },
+		'/v1/event': { 'event': 'pageview' },
+		'/v1/branchview': {}
+	},
+	allowErrorsInCallback: false
+};
+
+// Used by 3_api.js to determine whether a request should be blocked
+utils.userPreferences.shouldBlockRequest = function(url, requestData) {
```
yntema: I think it might be more clear to just declare this within the object above. I was a little confused when I saw this method used and looked for it in the object above and didn't see it.

yntema:
👍

PR #561, opened 2018-05-14
DEVEX-329 - Ability to create long links in .link() function

rubinsingh:
Allows the WebSDK to return a properly formatted long link incase there is poor network connectivity or when tracking is disabled

### DIFF
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -65,7 +65,7 @@ utils.generateDynamicBNCLink = function(branchKey, data) {
 	if (!branchKey && !data) {
 		return;
 	}
-	if (typeof data["data"] !== "string") {
+	if (!data["data"] || typeof data["data"] !== "string") {
 		data = utils.cleanLinkData(data);
 	}
 	var addKeyAndValueToUrl = function(fallbackUrl, tagName, tagData) {
@@ -299,7 +299,6 @@ utils.cleanLinkData = function(linkData) {
 				data = safejson.parse(data);
 			}
 			catch (e) {
-				console.log(e);
 				data = { '_bncNoEval': true };
 			}
 			break;
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -61,6 +61,42 @@ utils.userPreferences = {
 	}
 };
 
+utils.generateDynamicBNCLink = function(branchKey, data) {
+	if (!branchKey && !data) {
+		return;
+	}
+	if (typeof data["data"] !== "string") {
+		data = utils.cleanLinkData(data);
+	}
+	var addKeyAndValueToUrl = function(fallbackUrl, tagName, tagData) {
+		var first = fallbackUrl[fallbackUrl.length - 1] === "?";
+		var modifiedFallbackURL = first ? fallbackUrl + tagName : fallbackUrl + "&" + tagName;
+		modifiedFallbackURL += "=";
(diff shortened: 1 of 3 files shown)


src/6_branch.js
```diff
@@ -1061,8 +1061,14 @@ Branch.prototype['logEvent'] = wrap(callback_params.CALLBACK_ERR, function(done,
 /*** +TOC_HEADING &Deep Linking& ^ALL ***/
 /*** +TOC_ITEM #linkdata-callback &.link()& ^ALL ***/
 Branch.prototype['link'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done, data) {
+	var dataCopy = utils.merge({}, data);
+	var keyCopy = this.branch_key;
 	this._api(resources.link, utils.cleanLinkData(data), function(err, data) {
```
yntema: Since you will need the result of `utils.cleanLinkData(data)` here and in `generateDynamicBNCLink`, why don't you just pull this out into a variable above instead of making a copy of data.

src/1_utils.js
```diff
@@ -61,6 +61,42 @@ utils.userPreferences = {
 	}
 };
 
+utils.generateDynamicBNCLink = function(branchKey, data) {
+	if (!branchKey && !data) {
+		return;
+	}
+	if (!data["data"] || typeof data["data"] !== "string") {
+		data = utils.cleanLinkData(data);
+	}
```
yntema: You could just pass in the result of `utils.cleanLinkData(data)` instead of calling this function again.

src/1_utils.js
```diff
@@ -61,6 +61,42 @@ utils.userPreferences = {
...
+	var addKeyAndValueToUrl = function(fallbackUrl, tagName, tagData) {
+		var first = fallbackUrl[fallbackUrl.length - 1] === "?";
+		var modifiedFallbackURL = first ? fallbackUrl + tagName : fallbackUrl + "&" + tagName;
+		modifiedFallbackURL += "=";
+		return tagName === "data" ? modifiedFallbackURL + encodeURIComponent(utils.base64encode(tagData)) : modifiedFallbackURL + encodeURIComponent(tagData);
```
yntema: I would recommend passing in the base64encoded value into the function instead of handling here in order to simplify the function.

src/1_utils.js
```diff
@@ -61,6 +61,42 @@ utils.userPreferences = {
...
+
+	var fallbackUrl = config.link_service_endpoint + '/a/' + branchKey + '?';
+	var topLevelKeys = [ "tags", "alias", "channel", "feature", "stage", "campaign", "type", "duration", "sdk", "source", "data" ];
+	for (var i = 0; i < topLevelKeys.length; i++) {
+		var key = topLevelKeys[i];
+		var value = data[key];
+		if (value) {
+			if (key === "tags" && Array.isArray(value)) {
+				for (var index = 0; index < value.length; index++) {
+					fallbackUrl = addKeyAndValueToUrl(fallbackUrl, key, value[index]);
```
yntema: if you have multiple tags, won't it end up looking like this: `http://fallbackurl.com/?tags=tag1&tags=tag2&tags=tag3`? I don't think this will parse properly

src/1_utils.js
yntema: You should remove this conditional now since you are already passing in the result of the cleanLinkData function

yntema:
👍

PR #569, opened 2018-05-26
JOUR-1938 - Adds request round trip time tracking for WebSDK

rubinsingh:


src/3_api.js
```diff
@@ -325,6 +338,12 @@ Server.prototype.XHRRequest = function(url, data, method, storage, callback, nop
  */
 Server.prototype.request = function(resource, data, storage, callback) {
 	var self = this;
+	utils.currentRequestBrttTag = resource.endpoint + '-brtt';
+	if ((resource.endpoint === "/v1/url" || resource.endpoint === "/v1/has-app") && Object.keys(utils.instrumentation).length > 0) {
+		delete utils.instrumentation['-brtt'];
```
hemanth-s-a: These are only for link calls, right?

rubinsingh: We'll add the instrumentation object to /v1/has-app requests and to /v1/url requests.

rubinsingh:
@hemanth-s-a thanks for the initial review. I've updated the PR again based on tests that I performed today for retries and such.

PR #589, opened 2018-07-06
JOUR-2022 - Journeys Refactor

rubinsingh:


src/1_utils.js
```diff
@@ -353,7 +352,10 @@ utils.cleanLinkData = function(linkData) {
 		safejson.parse(data);
 	}
 	catch (e) {
-		data = goog.json.serialize(data);
+		// no need to serialize for v1/pageview call
```
ahmednawar: curious why `/v1/pageview` is different from other endpoints re serialization?

rubinsingh: @ahmednawar I've reverted this change in the subsequent commit to make it consistent with the other endpoints. Good call!

hemanth-s-a:
👍  Once circle passes.

hemanth-s-a:
👍

PR #600, opened 2018-08-01
Journey Dismissals

hemanth-s-a:
@rubinsingh @yntema @thomasgwozdz @ruthmpardee


src/branch_view.js
```diff
@@ -53,16 +51,19 @@ function renderHtmlBlob(parent, html, hasApp) {
...
+/**
+ * Checks if a journey should show based on dismiss time
+ * @param       {Object} branch
+ * @return      {boolean}
+ */
+function _areJourneysDismissedGlobally(branch) {
+	var globalDismissEndTimestamp = branch._storage.get('globalJourneysDismiss', true);
 
-	if (dismissTimeStamp === true || dismissTimeStamp > Date.now()) {
+	if (globalDismissEndTimestamp === true || globalDismissEndTimestamp > Date.now()) {
```
ruthmpardee: are there any time zone implications here? Will the `>` be able to handle if the `globalDismissEndTimestamp` is in a different time zone that `Date.now()` returns?

src/journeys_utils.js
```diff
@@ -449,22 +451,16 @@ journeys_utils.animateBannerEntrance = function(banner) {
...
  * @function journeys_utils.findDismissPeriod
- * @param {string} html
+ * @param {Object} metadata
  *
  * Checks template metadata to determine how long to not show banner when dismissed
  */
-journeys_utils.findDismissPeriod = function(html) {
+journeys_utils.findDismissPeriod = function(metadata) {
 	// default dismiss period to 1 week
 	var dismissPeriod = 7;
```
ruthmpardee: don't we want this to be in seconds now?

hemanth-s-a: This is a dead method. My bad. I'll remove this.

src/branch_view.js
hemanth-s-a: Timestamp doesn't have a timezone (well, I guess it does but its always in UTC).

https://github.com/BranchMetrics/web-branch-deep-linking/pull/600/files#diff-e9eb70faeef04133ebfef5782fe6aa2fR572 is where this value gets set. Value is always current timestamp + time the journey has to be dismissed in seconds (https://github.com/BranchMetrics/web-branch-deep-linking/pull/600/files#diff-e9eb70faeef04133ebfef5782fe6aa2fR472)

I'm glad we don't have to worry about timezone here :)

src/0_config.js
```diff
@@ -5,5 +5,5 @@ goog.provide('config');
 
 config.app_service_endpoint = 'https://app.link';
 config.link_service_endpoint = 'https://bnc.lt';
-config.api_endpoint = 'https://api.branch.io';
+config.api_endpoint = 'http://192.168.192.140:3002';
```
rubinsingh: @hemanth-s-a would you mind reverting changes to this file?

example.html
```diff
@@ -153,7 +153,7 @@ <h6>Who Gets the Points?</h6>
 		$('#init').click(function() {
 			request.html('branch.init();');
 			// Note that this example is using the key in two places, here and above
-			branch.init('key_live_feebgAAhbH9Tv85H5wLQhpdaefiZv5Dv', function(err, data) {
+			branch.init('key_live_abE0PFZNEDEv7paSxsskhllmzzjCsB67', function(err, data) {
```
rubinsingh: Also, let's revert changes to this file.

src/branch_view.js
```diff
@@ -170,11 +170,11 @@ branch_view.displayJourney = function(html, requestData, templateId, eventData,
 	document.body.removeChild(placeholder);
 
 	if (!utils.userPreferences.trackingDisabled && !testModeEnabled) {
-		branch_view.incrementAnalytics(eventData['branch_view_data']);
+		branch_view.incrementAnalytics(branchViewData);
```
rubinsingh: @hemanth-s-a it might make sense to remove the second pageview call in this PR

src/journeys_utils.js
```diff
@@ -500,26 +490,108 @@ journeys_utils.finalHookups = function(templateId, storage, cta, banner, hideBan
...
+	var globalDismissPeriod = !testModeEnabled
+		? journeys_utils._findGlobalDismissPeriod(metadata)
+		: 0;
+	journeys_utils.branch._publishEvent(eventName, journeys_utils.journeyLinkData);
+	journeys_utils.journeyDismissed = true;
+	journeys_utils.animateBannerExit(banner);
+	if (globalDismissPeriod !== undefined) {
+		storage.set('globalJourneysDismiss', globalDismissPeriod, true);
+	}
+	var journeyDismissals = journeys_utils._setJourneyDismiss(storage, templateId, audienceRuleId);
```
rubinsingh: journeysDismissals isn't used in _handleJourneyDismiss()

PR #617, opened 2018-11-14
[DEVEX-882]: added support to check if the Journey is displaying

code-crusher:
- Added _isJourneyDisplaying()_ method to check anytime if any Journey is displaying or not. 
- Added the test for the same also. 

**Reviewers:**
@rubinsingh @Sarkar


rubinsingh:
Closing PR because partner can use Journey event listeners to achieve use-case.

PR #622, opened 2018-12-11
DEVEX-926 INTENG-4362

aaaronlopez:


src/6_branch.js
```diff
@@ -462,6 +582,15 @@ Branch.prototype['init'] = wrap(
...
 								);
 							}
+							// TODO: check for Twitter referrer, try to use a decision from backend to make this change dynamic without web sdk
+							else if (!branchMatchIdFromOptions && utils.getParamValue('branchify_url') && self._referringLink()) {
+								var linkOptions = {
+									make_new_link: false,
+									open_app: true
+								};
+								this['branch']['deepview']({}, linkOptions);
+								journeys_utils.branch._publishEvent('willNotShowJourney');
```
ahmednawar: how can the backend avoid publishing a cta view in this case?

src/6_branch.js
```diff
@@ -462,6 +582,15 @@ Branch.prototype['init'] = wrap(
 									journeyInTestMode
 								);
 							}
+							// TODO: check for Twitter referrer, try to use a decision from backend to make this change dynamic without web sdk
+							else if (!branchMatchIdFromOptions && utils.getParamValue('branchify_url') && self._referringLink()) {
```
aaustin: Can you explain this logic?
- !branchMatchIdFromOptions: Just curious why?
- self._referringLink: Won't this not be referred?

src/6_branch.js
aaaronlopez: working on a way to send `auto_branchify=true` in the outgoing v1/deepview request and not published if true: https://github.com/BranchMetrics/branch-backend/pull/5097/files

src/6_branch.js
aaaronlopez: !branchMatchIdFromOptions: The thinking was if we have the branch_match_id, rely on it for matching.

self._referringLink: It will be set to the `branchify_url` after v1/open sets the link data on session storage. It's not really the same a `document.referrer`.

aaustin: Understood.

src/6_branch.js
```diff
@@ -508,7 +518,7 @@ Branch.prototype['init'] = wrap(
 			}
 		};
 
-		if (sessionData && sessionData['session_id'] && !link_identifier) {
+		if (sessionData && sessionData['session_id'] && !link_identifier && !utils.getParamValue('branchify_url')) {
```
aaaronlopez: this is to trigger a v1/open everytime there's a `branchif_url`

src/6_branch.js
```diff
@@ -1394,6 +1404,10 @@ Branch.prototype['deepview'] = wrap(callback_params.CALLBACK_ERR, function(done,
 
 	cleanedData['banner_options'] = options;
 
+	if (options['auto_branchify']) {
+		cleanedData['auto_branchify'] = true;
```
aaaronlopez: this is to not double count the CTA clicks

src/branch_view.js
```diff
@@ -225,6 +225,9 @@ branch_view._getPageviewRequestData = function(metadata, options, branch, isDism
 	if (linkClickId) {
 		obj.data['link_click_id'] = linkClickId;
 	}
+	if (sessionStorage['data'] && JSON.parse(sessionStorage['data'])['+referrer']) {
+		obj.data['+referrer'] = JSON.parse(sessionStorage['data'])['+referrer'];
+	}
```
aaaronlopez: we are very strict about what we send out to the API, so I need to add it to this request thing

ahmednawar:
LGTM but I defer to rubin for the final thumb

src/6_branch.js
rubinsingh: let's double check what happens when you have a _branch_match_id and a branchify_url as a query param, what would take precedence in this case? Would link data be returned from branch_match_id or from the branchify_url?

src/branch_view.js
rubinsingh: Can we make this simpler since we're doing JSON.parse() twice?

src/6_branch.js
aaaronlopez: confirmed `branch_match_id` takes precedence here. https://91szw-alopez.branchbeta.link/3vGThQ3XCS goes to http://192.168.192.251:8000/example.html?branchify_url=https%3A%2F%2F91szw-alopez.branchbeta.link%2Fnobranchlink&utm_medium=marketing&_branch_match_id=546068215899985938 but I have the `https://91szw-alopez.branchbeta.link/3vGThQ3XCS` only

rubinsingh:
👍 can we run regressions and branch micro against a build from this branch?

PR #629, opened 2019-01-12
DEVEX-857 - Expect JourneyLinkData to be an object in v1/pageview request 

rubinsingh:
Instead of having JourneyLinkData be embedded inside of the Journey HTML templates, it'll now be included as it's own key.


hemanth-s-a:
👍

PR #630, opened 2019-01-27
JOUR 2801 -  Adds journey_link_data to data returned in Journeys event handlers

rubinsingh:
cc. @aaaronlopez  @hemanth-s-a

### DIFF
--- a/src/journeys_utils.js
+++ b/src/journeys_utils.js
@@ -665,14 +665,15 @@ journeys_utils.animateBannerExit = function(banner, dismissedJourneyProgrammatic
 		}
         	journeys_utils.isJourneyDisplayed = false;
 	}, speedAndDelay);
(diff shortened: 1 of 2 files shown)


aaaronlopez:
👍 makes sense in general, but would like to understand the failure more clearly. 

Was it that the `journey_link_data` was getting wiped when `linkData={}`? Also, would like to understand why `data['journey_link_data'] = {};`

rubinsingh:
@aaaronlopez the issue here is that in the current released version of the WebSDK, CTA link data is embedded within a key called "journey_link_data" when it is returned in event handlers. When data is returned from the server, the CTA data is not embedded within this top level key so we have to build a new JSON object.

aaaronlopez:
👍 clarified

PR #635, opened 2019-04-03
[INTENG-4678]: add support for Safari IAB safe-area

code-crusher:
Description:
As of now, in the Safari IAB, the banner gets overlapped by the bottom nav bar. This change ensures it's always above it.

Reviewers:
@rubinsingh @aaaronlopez


aaaronlopez:
👍

PR #636, opened 2019-04-05
[DEVEX-960]: make safe property access for SafeAreaRequired

code-crusher:
The object property was not safe before, this change fixes it.

### DIFF
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -952,11 +952,11 @@ utils.removePropertiesFromObject = function(objectToModify, keysToRemove) {
 	}
 };
 
-utils.getPropertyValueByNameFromObject = function(object, name) {
-	if (object && typeof object === "object") {
-		for (var key in object) {
-			if (object.hasOwnProperty(key) && key===name) {
-				return object[key];
+utils.getPropertyValueByNameFromObj = function(objectToSearch, name) {
+	if (objectToSearch && typeof objectToSearch === "object") {
+		for (var key in objectToSearch) {
+			if (objectToSearch.hasOwnProperty(key) && key===name) {
+				return objectToSearch[key];
 			}
 		}
 	}
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -952,6 +952,16 @@ utils.removePropertiesFromObject = function(objectToModify, keysToRemove) {
 	}
 };
 
+utils.getPropertyValueByNameFromObject = function(object, name) {
+	if (object && typeof object === "object") {
+		for (var key in object) {
+			if (object.hasOwnProperty(key) && key===name) {
+				return object[key];
+			}
+		}
+	}
+}
+
 // v2/event utility functions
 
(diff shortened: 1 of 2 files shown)


GeneShay:
Tested - looks good, and respects twitter in-app browser as expected.

src/journeys_utils.js
```diff
@@ -439,7 +439,8 @@ journeys_utils.animateBannerEntrance = function(banner) {
 			banner.style.top = '0';
 		}
 		else if (journeys_utils.position === 'bottom') {
-			if(!journeys_utils.journeyLinkData.journey_link_data['safeAreaRequired']) {
+			var safeAreaRequired = utils.getPropertyValueByNameFromObj(journeys_utils.journeyLinkData['journey_link_data'], 'safeAreaRequired') || false;
```
aaaronlopez: Can we just do: 
```
if (journey_utils.journeyLinkData && journey_utils.journeyLinkData.journey_link_data && journeys_utils.journeyLinkData.journey_link_data['safeAreaRequired'])
```

If necessary to check the key exists, 

add `Object.keys(journeys_utils.journeyLinkData.journey_link_data).hasOwnProperty('safeAreaRequired')`?

src/1_utils.js
```diff
@@ -952,6 +952,16 @@ utils.removePropertiesFromObject = function(objectToModify, keysToRemove) {
 	}
 };
 
+utils.getPropertyValueByNameFromObj = function(objectToSearch, name) {
+	if (objectToSearch && typeof objectToSearch === "object") {
+		for (var key in objectToSearch) {
```
rubinsingh: A for-loop here seems un-necessary

src/journeys_utils.js
code-crusher: So the initial error which was coming in was because I was using `journeys_utils.journeyLinkData.journey_link_data` - whenever I use the build.min.js with this, it throws the undefined object error. 

The initial problem was not with the `safeAreaRequired` key but was with the above. 

I ran the test the build.js, so it worked there and I assumed it should be good with minify too, but this fails. Do you have any thoughts on why this could be failing? I can see, in other places also, it is referenced via `[]` and not `.`:  https://github.com/BranchMetrics/web-branch-deep-linking/blob/master/src/journeys_utils.js#L741

code-crusher: Additionally, if I go with `Object.keys(journeys_utils.journeyLinkData['journey_link_data']).hasOwnProperty('')` - this gives me false always, irrespective of the key I use, even with `journey_id`.

aaaronlopez: Gotcha, so if that's the case, does the below also fail through minification?

```
if (journey_utils.journeyLinkData && journey_utils.journeyLinkData['journey_link_data'] && journeys_utils.journeyLinkData['journey_link_data']['safeAreaRequired'])
```

aaaronlopez:
:thumbs-up:

aaaronlopez:
👍

PR #640, opened 2019-06-25
INTENG-6603: Start sending current_url on v1/open

aaaronlopez:
The API Stuff is in.


src/1_utils.js
```diff
@@ -1042,6 +1042,10 @@ utils.getInitialReferrer = function(referringLink) {
 	return document.referrer;
 };
 
+utils.getCurrentUrl = function() {
```
aaaronlopez: I put it in a helper just in case

PR #645, opened 2019-07-09
Exception when sending current_url on v1/open in iframe

grif-ecwid:
- Fix Same Origin Policy restriction issue

There is exception when init branch.io in iframe 
```
Uncaught DOMException: Blocked a frame with origin "example.com" from accessing a cross-origin frame.
```

### DIFF
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -1043,7 +1043,13 @@ utils.getInitialReferrer = function(referringLink) {
 };
 
 utils.getCurrentUrl = function() {
-	return window.top.location.href || "";
+	if (utils.isIframeAndFromSameOrigin()) {
+		return window.top.location.href || "";
+	} else if (utils.isIframe) {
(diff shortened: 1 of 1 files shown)


src/1_utils.js
```diff
@@ -1043,7 +1043,13 @@ utils.getInitialReferrer = function(referringLink) {
 };
 
 utils.getCurrentUrl = function() {
-	return window.top.location.href || "";
+	if (utils.isIframeAndFromSameOrigin()) {
+		return window.top.location.href || "";
+	} else if (utils.isIframe) {
```
aaaronlopez: This would return a function instead of executing it.

aaaronlopez:
@grif-ecwid thank you for your contribution! We will fix this soon.

aaaronlopez:
I will merge this PR and modify the final edit.

PR #672, opened 2020-01-30
Get rid of specifc branch_id validation and make servers responsible for validation

GeneShay:
This simply makes the verification for branch_ids to only check that it is a String. Services will be responsible for validation and processing of this instead, which will untie our hands in terms of what these should look like, as well as open the door for encryption. 

Tested this change manually with **v1/open** and **/has_app**. Using the following example as a **branch_session_first** example in local storage:

> {"session_id":"750887asdfasdfasdf405456214440","identity_id":"7508874054asdfasdf60858218","link":"https://appsolutely.app.link?%24identity_id=7508874054asdfasdf60858218","data":"{\"+clicked_branch_link\":false,\"+is_first_session\":true}","browser_fingerprint_id":"asdasdfasdfsadff750887405443342820","has_app":false,"link_click_id":"54354654351sadsdaf"}

Both return a 400.

### DIFF
--- a/src/2_resources.js
+++ b/src/2_resources.js
@@ -80,14 +80,12 @@ function validator(required, type) {
 	};
 }
 
-var branch_id = /^[0-9]{15,20}$/;
-
 function defaults(obj) {
 	var def = {
-		"browser_fingerprint_id": validator(true, branch_id),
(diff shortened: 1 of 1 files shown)


aaaronlopez:
👍

PR #679, opened 2020-02-27
[CORE-729] base64 encoding of BFP

GeneShay:
@rubinsingh @ahmednawar @aaaronlopez 
I think this should be sufficient for this functionality. 
I tested the following with these changes:
v1/open
v1/pageview
v1/event
v2/event/standard
v2/event/custom
/has-app

### DIFF
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -605,6 +605,19 @@ utils.base64encode = function(input) {
 	return output;
 };
 
+/**
+ * Check if a String is a BASE64 encoded value
+ * @param {string} str 
+ */
+utils.isBase64Encoded = function(str) {
+	if (typeof str != "string") { return false; }
+	if (str ==='' || str.trim() ==='') { return false; }
+    try {
+        return btoa(atob(str)) === str;
+    } catch (err) {
+        return false;
+    }
+} 
 
 /**
  * Add event listeners to elements, taking older browsers into account
--- a/src/2_session.js
+++ b/src/2_session.js
@@ -15,7 +15,8 @@ goog.require('storage');
 session.get = function(storage, first) {
 	var sessionString = first ? 'branch_session_first' : 'branch_session';
 	try {
-		return safejson.parse(storage.get(sessionString, first)) || null;
+		let data = safejson.parse(storage.get(sessionString, first)) || null;
+		return decodeBFPs(data);
 	}
 	catch (e) {
 		return null;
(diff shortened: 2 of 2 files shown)


src/2_session.js
```diff
@@ -43,6 +45,37 @@ session.update = function(storage, newData) {
...
 };
+
+/**
+ * Encodes BFP in data object with Base64 encoding.
+ * BFP is supposed to be Base64 encoded when stored in local storage/cookie.
+ * @param {Object} data 
+ */
+function encodeBFPs(data){
+	if(data && !utils.isBase64Encoded(data["browser_fingerprint_id"])){
+		data["browser_fingerprint_id"] = btoa(data["browser_fingerprint_id"]);
```
ahmednawar: was the decision to still keep the keys with the same name?

src/1_utils.js
```diff
@@ -605,6 +605,19 @@ utils.base64encode = function(input) {
 	return output;
 };
 
+/**
+ * Check if a String is a BASE64 encoded value
+ * @param {string} str 
+ */
+utils.isBase64Encoded = function(str){
+	if(typeof str != "string"){ return false; }
```
ahmednawar: what if str is null?

GeneShay: null is of type "object", or if you mean undefined - type of will return "undefined". So this function will correctly return false

src/2_session.js
GeneShay: On the email thread between Alex, Aaron, You and I - there is no mention about changing the key. My understanding is that since all of this is strictly for obfuscation reason than anything else - we were not planning on encrypting the key.

src/1_utils.js
ahmednawar: add space after `if`, space after `)`

src/1_utils.js
```diff
@@ -605,6 +605,19 @@ utils.base64encode = function(input) {
...
+	if (str ==='' || str.trim() ===''){ return false; }
```
ahmednawar: add space after `)`

src/2_session.js
```diff
@@ -43,6 +45,37 @@ session.update = function(storage, newData) {
...
+	return data;
+}
+
+/**
+ * Decodes BFPs in data object from Base64 encoding.
+ * BFP is supposed to be Base64 encoded when stored in local storage/cookie. 
+ * @param {Object} data
+ */
+function decodeBFPs(data){
+	if(data && utils.isBase64Encoded(data["browser_fingerprint_id"])){
```
ahmednawar: space after `if`, same to the following if as well

ahmednawar:
👍

PR #684, opened 2020-03-04
fix for alternate_browser_fingerprint_id encoding

GeneShay:
The issue was that when alternative_browser_fingerprint_id is set to undefined - it gets encoded. Previously it would just be undefined, but now "undefined" would just get encoded, which caused the issue.


aaaronlopez:
👍

PR #689, opened 2020-03-10
Core 828 fix decoding of BFPs when accessed from storage directly

GeneShay:
This was tested manually and automation tests in the JIRA that failed before now pass.

CC: @jwangler-branch @aaaronlopez @scott-branchio @khuisman-branch @adams-branch


aaaronlopez:
lgtm too 👍

PR #691, opened 2020-03-10
Merge of fingerprint encoding and developer identity 

jwangler-branch:
- Refactorings of session.patch()
- Adapt resources to developer identity


src/2_resources.js
```diff
@@ -129,8 +129,7 @@ resources.linkClick = {
 		"link_url": validator(true, validationTypes.STRING)
 	},
 	params: {
-		"click": validator(true, validationTypes.STRING),
-		"identity": validator(true, validationTypes.STRING)
```
aaaronlopez: The first bool arg to `validator` decides whether the parameter is required. Why not instead of removing do we not just set it to `false`?

src/2_session.js
```diff
@@ -57,15 +57,16 @@ session.update = function(storage, newData) {
  */
 session.patch = function(storage, data, updateLocalStorage){
 
-	const merge = (source, patch) =>{
-		return utils.merge(source, patch);
```
aaaronlopez: Is the ` utils.merge` broken? If so, can we not fix the utils function?

src/2_resources.js
aaaronlopez: Same with the other ones below. I am more actually curious, not actually sure if it's fair to just make it optional with validation. It just sounds better to me in theory.

src/2_session.js
jwangler-branch: 👍

aaaronlopez:
👍

PR #696, opened 2020-03-17
Allow customers to override Journeys CTA links - JOUR-3883

jwangler-branch:
- setBranchViewData() should allow the user to specify a $journeys_cta parameter. Witch overrides the branch link entirely

- If the user tries to update link redirects using existing Branch params (e.g. $android_url, $ios_url, $fallback_url) then we should allow that.


src/journeys_utils.js
```diff
@@ -908,3 +908,72 @@ journeys_utils.setJourneyLinkData = function(linkData) {
 	}
 	journeys_utils.journeyLinkData = data;
 };
+
+journeys_utils.hasInBranchViewData = function(fieldName) {
```
hemanth-s-a: What does the name of the method mean?

jwangler-branch: Ok, bad naming. :) The function returns **true** if in BranchViewData a field with the name: FieldName

jwangler-branch: a better name could be `hasBranchViewData(fieldName) or isFieldInBranchViewData(fieldName) or  doesParameterExitsInBranchViewData(parameterName) or exitsBranchViewData(name)`

jwangler-branch: `exitsValueInBranchViewData(name)`

rubinsingh:
@jwangler-branch could you please add build.js and build.min.js to the gitignore file? I'd also remove them from your PR.

dist/build.js
```diff
@@ -2799,11 +2843,15 @@ Branch.prototype.logEvent = wrap(callback_params.CALLBACK_ERR, function(a, b, c,
   });
 });
 Branch.prototype.link = wrap(callback_params.CALLBACK_ERR_DATA, function(a, b) {
-  var c = utils.cleanLinkData(b), d = this.branch_key;
+  var c = utils.cleanLinkData(b), c = journeys_utils.trySetBranchViewDataUrls(b), d = this.branch_key;
   this._api(resources.link, c, function(b, f) {
     if (b) {
       return a(b, utils.generateDynamicBNCLink(d, c));
     }
+    console.log(33);
```
rubinsingh: console.log()

src/journeys_utils.js
```diff
@@ -908,3 +908,72 @@ journeys_utils.setJourneyLinkData = function(linkData) {
...
+			}
+			return value;
+		}, data);
+	};
+
+	try {
+		var data = (JSON.parse(linkElements.data));
+		linkElements.data = JSON.stringify(assignUrls(data));
+	}catch(e){
+		console.log(e);
```
rubinsingh: Are we sure we want to print errors to the console?

src/journeys_utils.js
```diff
@@ -908,3 +908,72 @@ journeys_utils.setJourneyLinkData = function(linkData) {
...
+	 return undefined;
+};
+
+journeys_utils.getJourneyCtaLink = function () {
+	return journeys_utils.getBranchViewDataItemOrUndefined('$journeys_cta');
+};
+
+journeys_utils.trySetBranchViewDataUrls = function (linkElements, urls = ['$android_url', '$ios_url', '$fallback_url', '$desktop_url']) {
+	if(!linkElements){
+		console.log("setDefaultUrls - no data set");
```
rubinsingh: console.log()

src/journeys_utils.js
```diff
@@ -908,3 +908,72 @@ journeys_utils.setJourneyLinkData = function(linkData) {
...
+		return linkElements;
+	}
+
+	var assignUrls = function (data) {
+		return urls.reduce((value, url)=>{
```
rubinsingh: Does ES6 code get minified properly?

src/6_branch.js
```diff
@@ -1252,12 +1252,19 @@ Branch.prototype['logEvent'] = wrap(callback_params.CALLBACK_ERR, function(done,
 /*** +TOC_ITEM #linkdata-callback &.link()& ^ALL ***/
 Branch.prototype['link'] = wrap(callback_params.CALLBACK_ERR_DATA, function(done, data) {
 	var linkData = utils.cleanLinkData(data);
+	linkData = journeys_utils.trySetBranchViewDataUrls(data);
```
rubinsingh: Can we change the name of this function? I'm also confused as to why we're modifying the .link() function if the code is supposed to affect the Journeys CTA only?

src/journeys_utils.js
```diff
@@ -908,3 +908,72 @@ journeys_utils.setJourneyLinkData = function(linkData) {
 	}
 	journeys_utils.journeyLinkData = data;
 };
+
+journeys_utils.exitsValueInBranchViewData = function(name) {
```
rubinsingh: exitsValueInBranchViewData -> getValueForKeyInBranchViewData = function(key)

jwangler-branch: 👍

src/6_branch.js
jwangler-branch: 👍

src/journeys_utils.js
jwangler-branch: 👍

src/journeys_utils.js
jwangler-branch: In some way, we have to notify someone if there is an exception. We don't have a logging system for the SDK so only console log works. But I have an idea to deal with that. I'll try something 👍

dist/build.js
jwangler-branch: 👍

src/journeys_utils.js
jwangler-branch: Yes it works properly

PR #698, opened 2020-03-19
Allow customers to override Journeys CTA links - JOUR-3883

jwangler-branch:
- setBranchViewData() should allow the user to specify a $journeys_cta parameter. Witch overrides the branch link entirely

- If the user tries to update link redirects using existing Branch params (e.g. $android_url, $ios_url, $fallback_url) then we should allow that.


src/journeys_utils.js
```diff
@@ -908,3 +908,55 @@ journeys_utils.setJourneyLinkData = function(linkData) {
...
+		return journeys_utils.branch._branchViewData.data[name];
+	}
+	 return undefined;
+};
+
+journeys_utils.getJourneyCtaLink = function () {
+	return journeys_utils.getBranchViewDataItemOrUndefined('$journeys_cta');
+};
+
+journeys_utils.tryReplaceJourneyCtaLink = function (html){
```
rubinsingh: Can we call this replaceJourneyCtaLink(link, html), and we can check if link exists in the function:

if (!link) { return html; } 

... do other stuff

src/journeys_utils.js
```diff
@@ -908,3 +908,55 @@ journeys_utils.setJourneyLinkData = function(linkData) {
...
+
+journeys_utils.getBranchViewDataItemOrUndefined = function(name){
+	if(journeys_utils.getValueForKeyInBranchViewData(name)){
+		return journeys_utils.branch._branchViewData.data[name];
+	}
+	 return undefined;
+};
+
+journeys_utils.getJourneyCtaLink = function () {
+	return journeys_utils.getBranchViewDataItemOrUndefined('$journeys_cta');
```
rubinsingh: why not just call getValueForKeyInBranchViewData('$journeys_cta')?

src/journeys_utils.js
rubinsingh: or detectAndReplaceCustomCTALink()

PS this can be done in the backend as well:

https://github.com/BranchMetrics/branch-backend/blob/master/branch/branchview.js#L266

jwangler-branch: detectAndReplaceWithCustomCTALink. The function name shows us that we are not doing the Single Responsibility Principle in that function, which is not good.

jwangler-branch: But you are right a small refactoring could be an improvement.

src/journeys_utils.js
jwangler-branch: Try to increase the readability of the code.

src/0_config.js
```diff
@@ -5,5 +5,5 @@ goog.provide('config');
 
 config.app_service_endpoint = 'https://app.link';
 config.link_service_endpoint = 'https://bnc.lt';
-config.api_endpoint = 'https://api2.branch.io';
+config.api_endpoint = 'https://nightlyregression.api.beta.branch.io';
```
rubinsingh: This needs to remain 'https://api2.branch.io';

jwangler-branch: Monday morning

rubinsingh:
👍

PR #701, opened 2020-03-20
Fix lint errors and use vanilla JS

aaaronlopez:
cc @BranchMetrics/core-team 

I am fixing a bunch of lint errors and also moving some code back to old pre-ES6 rules. @rubinsingh and I aren't sure if we can safely move to ES6 yet, but I am opting to just push back to old vanilla JS for now to limit risk.


GeneShay:
👍

iosf45:
could you help me how I change my CNAME to custom.bnc.lt

PR #706, opened 2020-03-23
Allow customers to override Journeys CTA links

jwangler-branch:
* add journey CTA replacement variable ($journeys_cta)
* users are able to replace the journeys link by: `setBranchViewData({"$journeys_cta": "http://my.own.link"})`


rubinsingh:
👍

src/journeys_utils.js
```diff
@@ -908,3 +908,47 @@ journeys_utils.setJourneyLinkData = function(linkData) {
...
+
+	if(!journeys_utils.branch._branchViewData.data){
+		return false;
+	}
+
+	return journeys_utils.branch._branchViewData.data[key];
+};
+
+journeys_utils.getBranchViewDataItemOrUndefined = function(name){
+	if(journeys_utils.getValueForKeyInBranchViewData(name)){
```
aaaronlopez: hey @jwangler-branch , can you fix the spacing here? There is a linter in the package.json.

Just run `npm install` and it should pick up the hook going forward

aaaronlopez: for example: `if (journeys_utils.getValueForKeyInBranchViewData(name)) {`

similar patterns in the rest of the code

jwangler-branch: Yes, I'll set lint as auto-execute.. :)

PR #710, opened 2020-03-30
Has app fix falsify

jwangler-branch:
After we have initialized the SDK with `branch.init()` get in the callback the field **data**. This field contains among others **has_app**.  If there are no apps the value should be **false**. But in any case of **false** we got **null** as result. 

Solution:

The condition: `data['has_app'] || null,` gets always null or true, because we have falsified here. 
Se solution contains now a new function witch checks this case and provides the functionality to the SDK


rubinsingh:
👍

PR #716, opened 2020-05-14
[APPSEC-264] Escape SDK output to prevent XSS

jdee:
Escape HTML output from SDK open response before rendering to avoid XSS.

~~There doesn't seem to be a standard API for HTML escaping, at least from a quick search. It is undoubtedly also possible to do this with jQuery or bootstrap, which are both used in this example.~~ The [jQuery .text() method](https://api.jquery.com/text/#text2) makes this a one-liner.

~~I also need to find a way to test this, since it requires following a link. I think I can just hack my local copy of the SDK to open a malicious link on load.~~ I've verified this by adding HTML to the output, e.g.:

```js
			$('.info').text('<p>' + JSON.stringify(data) + '</p>');
```

Appreciate any feedback.


example.html
```diff
@@ -118,13 +118,11 @@ <h6>Who Gets the Points?</h6>
 
 		// Note that this example is using the key in two places, here and below
 		branch.init('key_live_feebgAAhbH9Tv85H5wLQhpdaefiZv5Dv', function(err, data) {
-			document.getElementsByClassName('info')[0].innerHTML = JSON.stringify(data);
+			// Escape any HTML in the output to avoid XSS.
+			$('.info').text(JSON.stringify(data));
 		});
 	</script>
 	<script type="text/javascript">
-		var info = $('.info');
```
jdee, line 125: These variables were completely unused.

jdee: Incorrect. The request and response fields were. Changing them to use ids instead of classes though, since they're each unique.

dyang-branch:
Thanks Jimmy, looks good for me. 👍

example.html
```diff
@@ -118,13 +118,13 @@ <h6>Who Gets the Points?</h6>
...
-			document.getElementsByClassName('info')[0].innerHTML = JSON.stringify(data);
+			// Escape any HTML in the output to avoid XSS.
+			$('#info').text(JSON.stringify(data));
 		});
 	</script>
 	<script type="text/javascript">
-		var info = $('.info');
-		var request = $('.request');
-		var response = $('.response');
+	  var request = $('#request');
```
aaaronlopez, line 126: only nitpick is the indent

PR #717, opened 2020-05-14
[INTENG-9351] Add customer_event_alias and content_items to custom events

jdee:
Not sure there were ever any tests for this or support for it in example.html.


jdee:
I'd really like to just add `content_items` to custom events too. This is supported elsewhere, e.g. iOS explicitly: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/master/Branch-SDK/Branch-SDK/BranchEvent.h#L68. I know RN supports it on top of iOS and Android.

jdee:
In fact, there's an example of content items with a custom event in the docs: https://help.branch.io/developers-hub/docs/web-advanced-features#track-events. Added that.

aaaronlopez:
Looks good. Do you think we should add any validation to src/2_resources.js for the key?

PR #718, opened 2020-05-14
[INTENG-9846] Missing identity & developer_identity in opens and pageviews

jdee:
Working on testing this. Again, I don't think there were any automated tests for this, but not entirely sure.


aaaronlopez:
I think we only need one or the other, not both. I think the key is whichever populates `user_data_developer_identity` in the eo_pageview and eo_web_session_start logs. See `routePageview` in `branch-backend/master/api/index.js`

jwangler-branch:
Hello Guys, I found this document: [here](https://branch.atlassian.net/wiki/spaces/PE/pages/494272513/Identity+scratchpad#Identityscratchpad-SDKs) At the end of this file you'll find little comments about developer_identity.  But I don't know whether this document up to date or not.

jdee:
Judging from https://github.com/BranchMetrics/web-branch-deep-linking-attribution/blob/master/src/6_branch.js#L184, I think for `/v1` endpoints, this is `identity`, and `developer_identity` for `/v2` endpoints. For pageviews and opens, this should be `identity`. Removed `developer_identity` from this change. Also made the parameter optional.

jdee:
I can verify this works using the `setIdentity` button in example.html. It includes `identity` in the payload of `/v1/open` and `/v1/pageview` requests afterward.

PR #720, opened 2020-05-19
Fix broken tests

jdee:
Unit tests have not been running in CI thanks to this: https://github.com/BranchMetrics/web-branch-deep-linking-attribution/blob/master/deployment/test.sh#L4. Running at the command line with mocha using `npm test` is troublesome because it uses Sauce Labs.

However, the test/test.html page runs unit tests in a browser. This fixes two failures from manual testing. Note there are a number of failures of tests of cookie-based storage when testing with Chrome, which doesn't seem to support writing cookies for file: URLs. All tests after this change pass in Firefox. I adjusted the failure message for each of these cases to warn that it is expected to fail in Chrome. Note that the last argument to any assertion is always an optional failure message, and it's usually unnecessary. For some reason, these tests supply a success message for just about every single assertion.

Also my first attempt at testing logEvent failed. I think I may need to stub out /v2 event posts properly. So for now I removed it.

That left two actual failing tests. In both cases, I've adjusted the test, assuming the code is correct. This needs review.

One is the test for whitelistSessionData, which takes in `developer_identity` and outputs both `identity` and `developer_identity`. See https://github.com/BranchMetrics/web-branch-deep-linking-attribution/blob/master/src/1_utils.js#L241-L242.

The other is a test for disableTracking(). The stored results include `"identity": null`. I'm not sure if this is a result of my recent change.

### DIFF
--- a/test/1_utils.js
+++ b/test/1_utils.js
@@ -57,21 +57,38 @@ describe('utils', function() {
 
 	describe('whiteListSessionData', function() {
 		it('should remove unwanted params', function() {
-			var data = {
+			/*
+			 * This actually describes how whitelistSessionData works. It looks for
+			 * developer_identity in the input and outputs both identity and
+			 * developer_identity. Perhaps this is used as a filter elsewhere.
+			 */
+			var input = {
+				"data": "string",
+				"data_parsed": {
+					"key": "value"
+				},
+				"has_app": true,
+				"developer_identity": "67890",
+				"referring_identity": "12345",
+				"referring_link": null,
+				"unwanted": "param"
+			};
+			var expected = {
 				"data": "string",
 				"data_parsed": {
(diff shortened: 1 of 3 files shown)


jdee:
Of the two actual failing tests that were fixed, both involve identity but were failing before my identity-related changes. So I'm not entirely sure whether the code should have changed or the tests. I'll review further but wouldn't mind a second opinion.

jdee:
The two tests that had to be adjusted both involved only the params returned to the calling script in the callback to `init` or `data`. For example, if `developer_identity` is present in an open response, both `identity` and `developer_identity` are returned with the same value to the calling script. I'm reluctant to change that sort of thing without further insight, so I think adjusting the test expectations is the right thing to do. Going to merge this.

PR #721, opened 2020-05-20
V2 event example

jdee:
There is an event-example.html here, but that just logs v1 events, mostly involving banners.

This exercises the `branch.logEvent()` function used to log v2 events. It first initializes the SDK and then will send a standard VIEW_ITEM event for a content item or a custom event you can type in. You can also supply a Customer Event Alias for both standard and custom events.

This is just a basic example to test some recent functionality.

Note that of the several example HTML files here, only example.html is hosted/deployed.

### DIFF
--- a/deployment/test.sh
+++ b/deployment/test.sh
@@ -1,5 +1,5 @@
 #/bin/bash
 
 make
-grunt test
+# grunt test
 git checkout .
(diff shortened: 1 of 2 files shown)


aaaronlopez:
👍

PR #722, opened 2020-05-20
[CORE-1058] remove identity from open

jdee:
This should not have been added. With this change. all unit tests still pass in Firefox. Need to add a test that would catch this.


aaaronlopez:
👍

PR #724, opened 2020-05-26
Restore identity in v1 open

jdee:
This restores the `identity` field in `/v1/open` posts, which was removed because of a failing regression test. This is necessary to populate developer identity for web session start events. The regression probably needs to be updated.


aaaronlopez:
👍

PR #727, opened 2020-06-18
Adds START_TRIAL standard event

ssg-luke:
Noticed I was getting an error that it was trying to send the START_TRIAL event as a custom event on the web SDK


jdee:
Thanks for catching that @ssg-luke ! We'll release this next week.

PR #729, opened 2020-06-23
[INTENG-9846] Null out developer identity on logout.

jdee:
Inspired by the handling in the iOS SDK:

https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/master/Branch-SDK/Networking/Requests/BranchLogoutRequest.m#L49-L51

The developer identity is nulled out. The identity ID is taken from the response payload. Note that identity is nulled out in the `data` object before saving to storage as well.


src/6_branch.js
```diff
@@ -834,7 +834,7 @@ Branch.prototype['logout'] = wrap(callback_params.CALLBACK_ERR, function(done) {
 		self.sessionLink = data['link'];
 		self.session_id = data['session_id'];
 		self.identity_id = data['identity_id'];
```
aaaronlopez, line 836: Wouldn't we want to drop `identity_id` too for conciseness? If it's stored in local storage, if we leave `identity_id` is it possible for it to be attributed on pageviews in addition to web session start?

aaaronlopez: I forget which one belongs to which. IIRC  identity_id -> v1/pageview, identity -> v1/open / web session start

jdee: The Android SDK, like the iOS SDK, also stores the `identity_id` received from the server. https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/master/Branch-SDK/src/main/java/io/branch/referral/ServerRequestLogout.java#L52

If memory serves we just added developer identity to /v1/open for the purpose of populating it in web session starts. That broke a regression, but we concluded it was the right thing to do and adjusted the regression. It was also added to pageviews for the same reason, though that didn't break a regression.

So that means that as long as `identity` is null, there will be no developer identity, even if `identity_id` is non-null. And finally, the session is retained through a logout/change of developer identity. Doesn't that get tied to an `identity_id`? I'm not sure.

jdee: Though after all of that, it looks like the server isn't returning a developer identity, so even without this code change, it should be nulled out. I'll need to dig further.
```json
{"session_id":"804487127261343793","identity_id":"804487127236518275","link":"https://bnctestbed.app.link/?%24identity_id=804487127236518275"}
```

jdee: Yeah, the `identity` is still being sent in pageviews and opens even with this change. More to come.

aaaronlopez: LMK if I can help with testing.

jdee:
Added a commit that removes `null` and `undefined` values from storage. Just using the `session.patch` method does the trick in this case, but it writes a bunch of `null` values into storage. I added a flag to `session.patch` and the underlying `utils.merge` called `removeNull` that deletes `null` values instead of inserting them.

Once this passes muster, I'd like to merge and run regressions.

src/1_utils.js
```diff
@@ -406,7 +407,14 @@ utils.merge = function(to, from) {
 
 	for (var attr in from) {
 		if (from.hasOwnProperty(attr)) {
-			to[attr] = from[attr];
+			var fromAttr = from[attr];
+			/* Only remove null and undefined, not all falsy values. */
+			if (removeNull && (fromAttr === undefined || fromAttr === null)) {
```
aaaronlopez, line 412: 👍

aaaronlopez:
👍 looks good, only this is I am not sure if it matters that we push the `dist/` folder but I guess it doesn't really

jdee:
I'd like to remove the dist folder from SCM, but I think some things broke when I first tried it. I also saw though that the deploy job regenerates it. We ought to straighten that out. There's no very good reason to check in things that are generated and confuse PRs.

PR #730, opened 2020-06-25
Fix broken deploy script

jdee:
Also fix example.html for the third time.

This is the commit that broke the last example.html deployment, message `Pushing release 2.54.1 [ci skip]`: https://github.com/BranchMetrics/web-branch-deep-linking-attribution/commit/f108ef850a3f062badfb96dc28f0754edc1e7cdc#diff-62ad28cc71ae9a54a611bc2d551533f6L121

That commit is generated here: https://github.com/BranchMetrics/web-branch-deep-linking-attribution/blob/master/deployment/deploy.sh#L84

At no point before that in the `.circleci/config.yml` or any script is example.html touched. I have a feeling somehow we may not be getting a pristine checkout, and a previous version is lying around uncommitted, so I added `git reset HEAD .; git checkout -- .` early in the process.

This may or may not be the solution, but it seems unlikely to make anything worse. If this doesn't fix the problem with this deploy, we can deploy the file manually (again).


deployment/deploy.sh
```diff
@@ -53,6 +53,12 @@ if [ "$CIRCLE_BRANCH" == 'production' ]; then
 
   echo -en "\n${GREEN}Extracted Changelog:\n$INSERT\n$CHANGELOG\n${NC}\n"
 
+  echo -en "\n${GREEN}Resetting local repo${NC}\n"
+  # Avoid scooping up any local changes left over from previous builds.
+  # These commands ensure a start from a pristine repo.
+  git reset HEAD .
```
jdee, line 59: These two commands should be harmless at least.

aaaronlopez:
👍

PR #732, opened 2020-07-06
[APPSEC-264] Correct template that generates example.html

jdee:
Having trouble testing this on Mac because of errors that occur when processing the template. These errors don't occur in CI/CD (Linux).

Remove git commands added that accomplished nothing and did not address the template issue.


aaaronlopez:
👍

PR #734, opened 2020-07-14
[INTENG-10184] Cannot call certain methods before init()

jdee:
All method calls are executed on a queue. This change allows `setBranchViewData` and `disableTracking` to be called before `init`. Currently, an error is generated. To test, add the following in the example.html in the local repo, before the call to `branch.init` (line 119):

```js
		branch.setBranchViewData({
			data: {
				'$ios_deeplink_path': 'abc/123'
			}
		}, function(err) {
			if (err) console.log('Error from setBranchViewData before init: ' + err);
		});
```

Before this change, the error is generated, and the parameter does not appear in the `/v1/pageview` payload. With this change, no error and the parameter is logged in the pageview.

Because generated code is included in the repo, this tiny change is not that clear. The third parameter to the `wrap` function at the beginning of `src/6_branch.js` is `init`, which is checked at line 100. This change just sets that value to `true` for these two functions and elides the `if (!init)` block.


dist/build.js
```diff
@@ -2939,7 +2939,7 @@ function _setBranchViewData(a, b, c) {
 }
 Branch.prototype.setBranchViewData = wrap(callback_params.CALLBACK_ERR, function(a, b) {
   _setBranchViewData.call(null, this, a, b);
-});
+}, !0);
```
aaaronlopez, line 2942: I was so thrown off by the `!0` then realized this was the compilation. I forget what we did last time, commit build.js or not, either way, I know this isn't you directly.

jdee: I guess the benefit of having this in the repo is that people who want to test with the local example.html don't have to install Clojure, Java, Python, compile, etc. It is generated by CD when deployed to prod and pushed back to the repo anyway. But I guess I'm not a fan of ever having to work from a local repo that isn't clean. You shouldn't have to avoid committing local changes that accompany a code change. I think I might prefer to remove these from the repo. If anyone wants to test with the local files, they can build. But we ought to have instructions for them in the repo.

aaaronlopez:
👍

PR #736, opened 2020-07-17
[INTENG-10570] Updated STANDARD Events

msiddiq-branch:
Updated these below events under BRANCH_STANDARD_EVENTS as these were tagged as standard but branch Web SDK would fire these events as the custom event leading to 400 error.

- CLICK_AD
- START_TRIAL
- VIEW_AD
- INVITE
- RESERVE

https://branch.atlassian.net/browse/INTENG-10570


code-crusher:
Thanks, @msiddiq-branch were you able to make v1/event calls with the new events to validate?

jdee:
I guess these are v2 events after all: https://help.branch.io/developers-hub/docs/tracking-commerce-content-lifecycle-and-custom-events, so we should go with this change. Thanks @msiddiq-branch!

code-crusher:
👍

jdee:
PS - I added this page to test v2 events: https://github.com/BranchMetrics/web-branch-deep-linking-attribution/blob/master/event-v2-example.html. I was thinking to add a pulldown to select all valid standard events, but that variable is kind of hidden. We could maybe expose it. Anyway, for now this should be a convenient way to test.

PR #737, opened 2020-07-22
[INTENG-9975] Add support for a _web_open_delay_ms query param

jdee:
See also https://github.com/BranchMetrics/branch-backend/pull/5849. Together these allow for a `$desktop_web_open_delay_ms` key that can be used to delay the POST to `/v1/open` from `branch.init()` when opening in a browser on the web. This can help alleviate a race condition with desktop apps.

### DIFF
--- a/src/6_branch.js
+++ b/src/6_branch.js
@@ -569,6 +569,9 @@ Branch.prototype['init'] = wrap(
 			self.identity = permData['identity'];
 		}
 
+		// Execute the /v1/open right away or after _web_open_delay_ms.
+		var open_delay = parseInt(utils.getParamValue('_web_open_delay_ms'), 10);
+
 		if (!utils.isSafari11OrGreater()) {
 			self._api(
 				resources._r,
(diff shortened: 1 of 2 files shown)


src/6_branch.js
```diff
@@ -569,6 +569,9 @@ Branch.prototype['init'] = wrap(
 			self.identity = permData['identity'];
 		}
 
+		// Execute the /v1/open right away or after _web_open_delay_ms.
+		var open_delay = parseInt(utils.getParamValue('_web_open_delay_ms'), 10);
```
aaaronlopez, line 573: Curious why you left it open to be a pattern match instead of specific one?

jdee: I'm not sure what you mean. Do you mean the way a Regexp is used in the implementation of `utils.getParamValue`? I guess it could be made `utils.getParamValue('[?&]_web_open_delay_ms')`. This was basically modeled on the way the match ID is extracted using `utils.getParamValue('_branch_match_id')`. The idea though is that if the key is missing or otherwise can't be converted to a positive integer, it remains `undefined`, which short-circuits `utils.delay` for immediate (unqueued) execution.

aaaronlopez: I guess what I wanted to really know was why we wanted to have the platform name as the first of the key name. I was initially thinking it'd be platform agnostic (just because it can be). So like `$web_open_delay_ms` instead of expecting the platform prepended, `$desktop_web_open_delay_ms` and in theory, it could be anything prepended, right? `blah_web_open_delay_ms` based on this function.

Not a big deal either way.

jdee: Yeah. I guess I wanted this to be desktop-specific at the param level, hence `$desktop_web_open_delay_ms` to indicate it doesn't delay the open on redirect into an app, only on the web, and only on the desktop. But then the redirect param is implicitly web, so it could be `_open_delay_ms` instead of `_web_open_delay_ms`. I hadn't realized this function was using a regexp, so I can make it more specific. But open to alternative names, etc.

jdee: Updated here and in the backend to `_open_delay_ms`. Also checking now for the exact key, not a regexp.

aaaronlopez:
lgtm 👍 feel free to merge if the solution is greenlit by customer

PR #739, opened 2020-07-22
[minor] - INTENG-9671 -  Journeys Additional Margin Getting Added for Single Page App

jwangler-branch:
What's done?

- Added the variable `exitAnimationIsRunning` which indicates the run state of the exit animation. 

Why? 

If you display a journey the `body.style.marginTop` or `maringBottom` will be calculated. If the closing animation and the journey display function are running at the same time, it generates a kind of offset to that margins.

image

 That only happens if the host application executes `window.branch.closeJourney(callback);` and does a route change at the same time.

 For more details feel free to have a look a the Jira or ask me any time.


src/journeys_utils.js
```diff
@@ -894,7 +901,9 @@ journeys_utils.animateBannerExit = function(banner, dismissedJourneyProgrammatic
 		if (!dismissedJourneyProgrammatically) {
 			journeys_utils.branch._publishEvent('branch_internal_event_didCloseJourney', journeys_utils.journeyLinkData);
 		}
+
 		journeys_utils.isJourneyDisplayed = false;
+		setTimeout(function(){ journeys_utils.exitAnimationIsRunning = false; }, journeys_utils.animationSpeed )
```
jdee, line 906: Curious why this uses setTimeout when this is already inside a setTimeout. Seems like you could just assign the variable here.

jwangler-branch: This is the problem itself. I use a timeout because we using and CSS animation to adjust the margins. This animation has to be finished. Unfortunately, this happens after the code block is processed. Therefore I have to use this timeout 👍

PR #742, opened 2020-07-27
Add tests for utils.delay()

jdee:
These can be run in a browser by opening `test/test.html`. The suite tends to do best in Firefox.


jdee:
Also verified these pass on Windows in Microsoft Edge.

PR #744, opened 2020-08-05
[DASH-3022] - Change $journeys_cta to use original URL rather than Branch link

jwangler-branch:
What's done?

- if you set in `setBranchViewData` the variable $journeys_cta the branch link will be replaced by the variable URL. 

How to test?

If you sell `setBranchViewData({"$journeys_cta": "http://my.own.link"})` you should navigated to the target by clicking the banner.


jwangler-branch:
@jdee I added some unit tests

PR #746, opened 2020-08-06
[CORE-1134] Stop calling /_r in WKWebView on iOS

jdee:
This extends the handling for Safari 11+ to WKWebView on iOS.


jdee:
There's another call to `isSafari11OrGreater` here: https://github.com/BranchMetrics/web-branch-deep-linking-attribution/blob/master/src/3_api.js#L223. I would guess that also needs to change. Not entirely sure of the logic yet.

jwangler-branch:
> e's another call to `isSafari11OrGre`

I do not really sure what this call is doing

PR #750, opened 2020-08-11
Comment out broken/unused slack notification

jdee:
This was going to the archived #web-sdk channel anyway. It's the last thing that happens in the script, so failure is benign. We can restore this at need.


aaaronlopez:
👍

PR #755, opened 2020-10-15
INTENG-11116: Fix unsafe eval error

aaaronlopez:
Undoing usage of `goog.json.parse` and replacing with `safejson.parse` to prevent "'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive"

This was the original commit: https://github.com/BranchMetrics/web-branch-deep-linking-attribution/commit/35296af3d388779a75be748f3aa5b1684e313b5c#diff-d744691a59bf20af6d8cbb5a3dace22c


jdee (approved):
:+1:

jwangler-branch (approved):

PR #759, opened 2020-10-29
[INTENG-11098][CORE-1417] Identity fixes

jdee:
The reported issue is that `/v1/open` returns 500 when the developer identity includes a space. The `identity` field should not be sent in an open request in the first place. It was added erroneously some time back to make developer identity show up in web session start events. The real problem is that the `identity_id` was not being picked up from storage at initialization, so a new `identity_id` was generated. With this change developer identity shows up in web session start events, even when it includes whitespace.

Related changes:
- `/v1/profile` returns a new identity_id (same session_id). This is now updated in storage.
- `/v1/pageview` ignores the identity field but requires the identity_id field.
- Return the `identity` from the open response in the init callback. This was always null.
- Added a button to the example.html template to exercise `branch.first()`. As a result of not storing the identity_id correctly, this method was not working.

@BranchMetrics/core-team

### DIFF
--- a/src/6_branch.js
+++ b/src/6_branch.js
@@ -341,6 +341,11 @@ Branch.prototype['init'] = wrap(
 			utils.cleanApplicationAndSessionStorage(self);
 		}
 
+		// initialize identity_id from storage
+		// note the previous line scrubs this if tracking disabled.
+		var localData = session.get(self._storage, /* use local if possible */ true);
+		self.identity_id = localData && localData['identity_id'];
+
 		var setBranchValues = function(data) {
 			if (data['link_click_id']) {
 				self.link_click_id = data['link_click_id'].toString();
--- a/src/6_branch.js
+++ b/src/6_branch.js
@@ -343,7 +343,7 @@ Branch.prototype['init'] = wrap(
 
 		// initialize identity_id from storage
 		// note the previous line scrubs this if tracking disabled.
-		var localData = session.get(self._storage, /* use local if possible */ true);
+		var localData = session.get(self._storage, true);
 		self.identity_id = localData && localData['identity_id'];
 
 		var setBranchValues = function(data) {
--- a/src/6_branch.js
+++ b/src/6_branch.js
@@ -788,7 +788,8 @@ Branch.prototype['setIdentity'] = wrap(callback_params.CALLBACK_ERR_DATA, functi
 				safejson.parse(data['referring_data']) :
 				null;
 
-			session.patch(self._storage, { "identity": identity }, true);
+			// /v1/profile will return a new identity_id, but the same session_id
+			session.patch(self._storage, { "identity": identity, "identity_id": self.identity_id }, true);
 			done(null, data);
 		}
 	);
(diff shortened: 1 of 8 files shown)


jdee:
This is deployed at https://jdee.github.io/dist/build[.min].js, e.g. on https://jdee.github.io/example.html (pointing to prod).

src/6_branch.js
```diff
@@ -341,6 +341,11 @@ Branch.prototype['init'] = wrap(
 			utils.cleanApplicationAndSessionStorage(self);
 		}
 
+		// initialize identity_id from storage
+		// note the previous line scrubs this if tracking disabled.
+		var localData = session.get(self._storage, true);
```
aaaronlopez, line 346: The second param, bool, leads us to retrieve this from the `branch_session_first`. What happens if I want to change the identity for a session? i.e. first time I call `setIdentity("identityA")`. What if two days latest I want to call ``setIdentity("identityB")` which should yield a different (new?) `identity_id`.

Would we send the new `identityB`?

jdee: Yes. There is in fact a unit test for it added in this PR. On success, `setIdentity` updates local storage with the new identity and identity_id.

test/6_branch.js
```diff
@@ -607,11 +607,20 @@ describe('Branch', function() {
 			requests[0].callback(null, expectedResponse);
 			assert.deepEqual(requests[0].obj, expectedRequest, 'All params sent');
 		});
+
+		it('should update identity and identity_id in local storage', function(done) {
+			var branch = initBranch(true);
+			var assert = testUtils.plan(2, done);
+			branch.setIdentity('12345678', function(err, data) {
+				var localData = safejson.parse(localStorage.getItem('branch_session_first'));
+				assert.strictEqual(localData['identity'], '12345678');
```
jdee, line 616: @aaaronlopez This ensures that we pick up the right identity & identity_id at initialization.

src/6_branch.js
```diff
@@ -788,7 +793,8 @@ Branch.prototype['setIdentity'] = wrap(callback_params.CALLBACK_ERR_DATA, functi
 				safejson.parse(data['referring_data']) :
 				null;
 
-			session.patch(self._storage, { "identity": identity }, true);
+			// /v1/profile will return a new identity_id, but the same session_id
```
jdee, line 796: We were already updating identity in local storage. This change updates identity_id as well in local storage. That ensures that we get the right identity_id at initialization, to send in /v1/open.

jdee:
The change to return a non-null `identity` isn't working in testing, so I'll fix that.

jdee:
That's working. May have just taken time to propagate the updated SDK. I also updated whiteListSessionData to send data.identity as developer_identity. The value of data.identity is returned as developer_identity in the callback to setIdentity. These values were both the same before, populated from a field that will never exist. So now they're both the same but will have a meaningful value.

ahmednawar:
Looks ok to me but I will defer to Aaron's since he is more familiar with the web sdk code.

src/6_branch.js
aaaronlopez: ahh I see

aaaronlopez:
👍 looks good, thanks for the tests!

test/6_branch.js
aaaronlopez: 👍

PR #761, opened 2020-11-17
[INTENG-11551] Allow short-link generation when tracking disabled

jdee:
This just adds `/v1/url` to the endpoints that can be called with tracking disabled. It also includes a fix for a broken test after a recent change (the identity-related update). https://jdee.github.io/example.html has this change and a button to disable tracking. https://jdee.github.io/example-prod.html is the same thing pointing at prod.

SMS (INTENG-11512) is a little more involved, but also not too hard.

@BranchMetrics/core-team


jleebranch:
This looks good to me 👍

aaaronlopez:
👍

PR #762, opened 2020-11-17
[INTENG-11512] Allow SMS with tracking disabled

jdee:
This is a little more involved than the `/v1/url` fix. This change is deployed at https://jdee.github.io/example.html.

@BranchMetrics/core-team

### DIFF
--- a/src/1_utils.js
+++ b/src/1_utils.js
@@ -54,12 +54,32 @@ utils.userPreferences = {
 		// Used by 3_api.js to determine whether a request should be blocked
 		var urlParser = document.createElement('a');
 		urlParser.href = url;
+
+		// INTENG-11512
+		// To allow SMS when tracking disabled, we must allow GET <actual link>.
+		// This precludes a filter on the path. Only apply the whitelist to
+		// service endpoints.
+		var whiteListDomains = [ config.api_endpoint, config.app_service_endpoint, config.link_service_endpoint ];
+		// https://www.w3schools.com/jsref/prop_anchor_origin.asp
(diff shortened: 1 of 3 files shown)


src/1_utils.js
```diff
@@ -54,6 +54,21 @@ utils.userPreferences = {
 		// Used by 3_api.js to determine whether a request should be blocked
 		var urlParser = document.createElement('a');
 		urlParser.href = url;
+
+		// INTENG-11512
+		// To allow SMS when tracking disabled, we must allow GET <actual link>.
+		// This precludes a filter on the path. Only apply the whitelist to
+		// service endpoints.
+		var whiteListDomains = [ config.api_endpoint, config.app_service_endpoint, config.link_service_endpoint ];
```
aaaronlopez, line 62: there's already some of this logic for the api and app service here https://github.com/BranchMetrics/web-branch-deep-linking-attribution/blob/cdfcdc9a1c2fa743f257e432429a3b0b8d19b6db/src/1_utils.js#L46

and it's used on line 87 `var whiteListedEndpointWithData = utils.userPreferences.whiteListedEndpointsWithData[urlPath];`

can we use the same object and add the link service endpoint and check the path in that manner that way the `c` only gets hardcoded in one place?

aaaronlopez: ahhh but the issue is we need a domain check... I see,  would simply adding `/c/` there not work?

jdee:
@aaaronlopez As discussed, simplified the path check for SMS.

PR #770, opened 2021-02-09
CORE-1613: Adding support for v1/qr-code

aaaronlopez:
TODO:
- [x] Create response object with raw data and base64 options


aaaronlopez:
@jdee let me know what you think of this. I want to still add the object you suggested with at least the `raw` and `base64` results.

src/6_branch.js
```diff
@@ -1419,6 +1419,30 @@ Branch.prototype['sendSMS'] = wrap(
...
+		data['qr_code_settings'] = safejson.stringify(utils.convertObjectValuesToString(qrCodeSettings || {}));
+		this._api(
+			resources.qrCode,
+			utils.cleanLinkData(linkData),
+			function(error, rawBuffer) {
+				var qrCodeResult = {};
+				if (!error && rawBuffer) {
+					qrCodeResult.rawBuffer = rawBuffer;
+					try {
+						qrCodeResult.base64 = btoa(String.fromCharCode.apply(null, new Uint8Array(rawBuffer)));
```
aaaronlopez, line 1435: These should be fine with IE 10+ even `Unit8Array` constructor, but I'll double-check

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa

src/3_api.js
```diff
@@ -281,14 +281,20 @@ Server.prototype.jsonpRequest = function(requestURL, requestData, requestMethod,
  * @param {utils._httpMethod} method
  * @param {storage} storage
  * @param {function(?Error,*=,?=)=} callback
- * @param {?boolean=} noparse
+ * @param {?boolean=} noParse - _optional_ -
+ * @param {?string} reponseType - _optional_ -
```
jleebranch, line 285:
```suggestion
 * @param {?string} responseType - _optional_ -
```

src/3_api.js
```diff
@@ -301,8 +307,8 @@ Server.prototype.XHRRequest = function(url, data, method, storage, callback, nop
 		if (req.readyState === 4) {
 			utils.addPropertyIfNotNull(utils.instrumentation, brttTag, utils.calculateBrtt(brtt));
 			if (req.status === 200) {
-				if (noparse) {
-					data = req.responseText;
+				if (noParse) {
+					data = req.response;
```
jleebranch, line 311: Can you explain this change from 'responseText' to 'response'?

src/6_branch.js
jdee: This is nice. The only thing you might consider before releasing this new API is making these into methods like `getBase64()` or just `base64()`. That could allow you to defer the base64-encoding until it's requested and provide more opportunity to extend in the future, like adding optional params. Just a suggestion/thought.

jdee (approved):
👏

src/6_branch.js
```diff
@@ -1419,6 +1419,30 @@ Branch.prototype['sendSMS'] = wrap(
 	}
 );
 
+Branch.prototype['qrCode'] = wrap(
```
jleebranch, line 1422: Can you add examples/usage guide as we have for the other SDK methods?

jleebranch:
It looks good. And you already wrote the b-m tests for the web sdk so 🥳 . Just left a few comments on minor things.

src/6_branch.js
aaaronlopez: that makes sense! I agree with return methods that way we do defer the base64 encoding

src/3_api.js
aaaronlopez: added explanation

src/6_branch.js
aaaronlopez: added as a function. LMK if you have a better suggestion.

src/6_branch.js
```diff
@@ -1419,6 +1419,88 @@ Branch.prototype['sendSMS'] = wrap(
...
+			resources.qrCode,
+			utils.cleanLinkData(linkData),
+			function(error, rawBuffer) {
+				function QrCode() { };
+
+				QrCode.rawBuffer = rawBuffer;
+				QrCode.base64 = function() {
+					// First Encode array buffer as UTF-8 String, then Base64 Encode
+					if (this.rawBuffer) {
+						return btoa(String.fromCharCode.apply(null, new Uint8Array(this.rawBuffer)));
```
aaaronlopez, line 1495: I preferred to not "tryCatch" this so the errors are visible now that it's simply a function and we throw an error below too

jleebranch:
👍

PR #777, opened 2021-04-20
Update rollbar.py to use python3

yli-branch:
rollbar.py currently uses python2, which doesn't have request library installed. This meant deploy step was failing at running deployment/rollbar.py step. This PR fixes rollbar.py to use python3, and updates print syntax to python3 syntax.

### DIFF
--- a/deployment/rollbar.py
+++ b/deployment/rollbar.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import os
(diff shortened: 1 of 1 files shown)


aaaronlopez:
I heard from Rex/Jimmy that there may be issues with some other node dependencies with python3 but worth a shot 👍

aaaronlopez:
Check with infra, we may as well try if you think this will help

PR #779, opened 2021-04-29
[JRNYS-2][INTENG-11986] Avoid document.write()

jdee:
This does not directly impact page rating by Google PageSpeed, but it eliminates diagnostic warnings about `document.write()`. The SDK has been just constructing a web page as text in memory and then shoving it into an iframe. This change modifies the DOM elements on the iframe directly instead, collapsing two utility functions into one.

This change is preliminary but has been deployed at https://jdee.github.io/pure-journeys-demo. [This page](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fjdee.github.io%2Fpure-journeys-demo) no longer shows complaints about document.write().

There is a similar place in a code path called by `branch.banner()` that ~~could be~~ has been addressed in the same way but does not affect Google PageSpeed diagnostics. Unit tests ought to be provided for this. All existing unit tests still pass with these changes.

@BranchMetrics/core-team

### DIFF
--- a/src/4_banner_html.js
+++ b/src/4_banner_html.js
@@ -157,7 +157,11 @@ banner_html.iframe = function(options, action) {
 	var iframedoc = iframe.contentDocument || iframe.contentWindow.document;
 	iframedoc.head = iframedoc.createElement('head');
 	iframedoc.body = iframedoc.createElement('body');
-	iframedoc.body.innerHTML = html;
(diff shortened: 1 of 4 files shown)


jwangler-branch:
👍

aaaronlopez:
nice! lgtm 👍

PR #782, opened 2021-05-13
INTENG-12729: Sticky Interstitial Full Page

wwest-branch:
Adding overflow:hidden styling to .branch-banner-no-scroll, which is removed when user clicks to remove interstitial.

https://branch.atlassian.net/browse/INTENG-12729


jdee (commented):
Looks good, but we should add unit tests. You can run them in a browser by opening test/test.html.

wwest-branch:
> Looks good, but we should add unit tests. You can run them in a browser by opening test/test.html.

After looking through this with Joerg, we concluded that the animateBannerExit function would need to be heavily refactored for us to unit test this code change. As I mentioned to you earlier, the setTimeout inside of animateBannerExit never gets invoked inside of the unit test, and there are a lot of cases inside of animateBannerExit that could be preventing this from happening. That said, we do have a branch-micro test plan and are writing integration tests for this. @jwangler-branch

wwest-branch:
> Looks good, but we should add unit tests. You can run them in a browser by opening test/test.html.

Please disregard the comment above, I was able to add some unit tests.

jwangler-branch:
Ah, you figured it out 👍

wwest-branch:
> Ah, you figured it out 👍

Yea, @yli-branch  pointed me to using sinon fake timers which was able to get past the setTimeout. However, you are right in that there is a lot of other logic that we needed to account for, which made this a lot harder. A refactor breaking animateBannerExit into smaller functions will be a good project in the future.

jdee (approved):
Looks good. We can address further refactoring later. Thanks.

PR #810, opened 2022-06-30
BUX-931 Sets up background dismissal

yli-branch:
A new feature of the new center overlay and partial interstitial templates is in the introduction of a semi-opaque background, which will dismiss the journey on click and on scroll. This PR sets up the dismissal behavior, and removes some hardcoded values from a previous implementation of center overlay.


src/journeys_utils.js
```diff
@@ -625,8 +612,10 @@ journeys_utils.finalHookups = function(templateId, audienceRuleId, storage, cta,
...
-	journeys_utils._setupDismissBehavior('.branch-banner-close', 'didClickJourneyClose', storage, banner, templateId, audienceRuleId, metadata, testModeEnabled, branch_view);
+	journeys_utils._setupDismissBehavior('.branch-banner-continue', 'didClickJourneyContinue', storage, banner, templateId, audienceRuleId, metadata, testModeEnabled, branch_view, 'click');
+	journeys_utils._setupDismissBehavior('.branch-banner-close', 'didClickJourneyClose', storage, banner, templateId, audienceRuleId, metadata, testModeEnabled, branch_view, 'click');
+	journeys_utils._setupDismissBehavior('.branch-banner-dismiss-background', 'didClickJourneyBackgroundDismiss', storage, banner, templateId, audienceRuleId, metadata, testModeEnabled, branch_view, 'click'); // TODO add new dismissal_source
+	journeys_utils._setupDismissBehavior('.branch-banner-dismiss-background', 'didScrollJourneyBackgroundDismiss', storage, banner, templateId, audienceRuleId, metadata, testModeEnabled, branch_view, 'touchmove'); // TODO add new dismissal_source
```
ezoller-branch, line 618: Should the TODO comments be removed and move into a Jira ticket?

ezoller-branch (commented):
Nice job! Is there a reason why the files inside `dist` is being published to GitHub? I thought build files should stay local and be ignored in a `.gitignore` file.

yli-branch:
@ezoller-branch that's a good callout. You are right that dist is getting rebuilt in the release.sh script. I don't know if there's good reason to not it be gitignored. Flagging this to you @nsingh-branch in case it's already on your radar

src/5_banner.js
```diff
@@ -228,6 +228,18 @@ banner = function(branch, options, linkData, storage) {
 			};
 		}
 
+		var modalBackground = doc.getElementById('branch-banner-modal-background');
+
+		if (modalBackground) {
+			modalBackground.onclick = function(ev) {
```
JagadeeshKaricherla-branch, line 234: avoid duplicating the code https://github.com/BranchMetrics/web-branch-deep-linking-attribution/blob/4835ce92e197b314baa4857e96de49e72edd1c69/src/5_banner.js#L222 

Have a common event handler function like 

```
function closeBanner(ev) {
				ev.preventDefault();
				branch._publishEvent('willCloseBanner');
				closeBanner({}, function() {
					branch._publishEvent('didCloseBanner');
				});
			};
closeButton.addEventListener('click',closeBanner);
modalBackground.addEventListener('click',closeBanner);
```

yli-branch: @nsingh-branch would you want this new function live top-level, or should it be defined inside of other functions? (in this case, this code logic is used within finalHookupsCallback, so I could have it live within finalHookupsCallback, or within the banner function)

JagadeeshKaricherla-branch: @yli-branch Yes within finalHookupsCallback like my above code.
btw you are tagging a wrong reviewer here. ( nsingh-branch) is not the reviewer here.

src/journeys_utils.js
JagadeeshKaricherla-branch: @yli-branch : I think if click, touchmove are both attached to the same element(branch-banner-dismiss-background) then on a touch device - the order will be touch event handler ( touch move)  first followed by mouse events ( click). Do you need to trigger the same event handler function ( _setupDismissBehavior)  twice for a touch device tap action ( once for touchmove and other time for click) ?

- You can use preventDefault in  event handler so that subsequent click/mouse event can be prevented from calling.
- Other way is to  check  if its a touchscreen device or not and invoke one of the handlers but this can fail on edge cases like touch screen laptops... so not a good way 
if(('ontouchmove' in window){
..
}
else
{
}

src/5_banner.js
yli-branch: thanks for the callout!

src/journeys_utils.js
yli-branch: @JagadeeshKaricherla-branch The product requirement actually has me targeting two separate use cases on the same touch device. When I tested this code on a mobile device simulation on chrome browser, only one call to _handleJourneyDismiss is made per interaction, and it was able to distinguish between a touchmove (scroll) and a click. Should I expect it to be different on a real device?

PR #1138, opened 2026-06-10
fix(journeys): guard empty journey_link_data in setJourneyLinkData (EMT-3772)

wpinho-branch:
## Summary

`journeys_utils.setJourneyLinkData(linkData)` read `data['journey_link_data']['type']` (and `['variant']`) **unconditionally**, but only assigned `data['journey_link_data']` when `linkData` was a **non-empty** object. When `journey_link_data` was `{}` (or absent), `data['journey_link_data']` was `undefined` and the read threw `TypeError: Cannot read properties of undefined (reading 'type')`, aborting the Journey/banner render.

In production `journey_link_data` is normally populated, so it does not surface for typical Journeys — but any path delivering an empty value breaks the banner.

## Fix

Default the object before the reads in `src/journeys_utils.js`:

```js
data['journey_link_data'] = data['journey_link_data'] || {};
```

Normal behavior (populated data) is unchanged; for the empty/absent case, `journeyType`/`journeyVariant` resolve to `null` and `isDesktopJourney` to `false`.

## Tests

Added a `setJourneyLinkData` describe block in `test/journeys_utils.js` covering:
- empty object `{}` (the regression)
- `undefined`
- populated `linkData` (normal path)

Verified the tests are genuine regression tests — they fail without the fix with the exact `TypeError` from the ticket, and pass with it.

- `make` (Closure build): 0 errors
- `npm test`: 158 passing

## Ticket

[EMT-3772](https://branch.atlassian.net/browse/EMT-3772)

## Out of scope, noticed in passing

`.mocharc.js` only runs 5 of the 10 test files; ~146 tests in `2_storage.js`, `3_api.js`, `6_branch.js`, `7_integration.js`, and `journeys.js` are not executed by `npm test`/CI. Worth a separate ticket to triage/re-enable — not addressed here to keep this PR surgical.

[EMT-3772]: https://branch.atlassian.net/browse/EMT-3772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/journeys_utils.js
+++ b/src/journeys_utils.js
@@ -1027,6 +1027,7 @@ journeys_utils.setJourneyLinkData = function(linkData) {
 		utils.merge(data['journey_link_data'], linkData);
 	}
 	journeys_utils.journeyLinkData = data;
+	data['journey_link_data'] = data['journey_link_data'] || {};
 	journeys_utils.journeyType = data['journey_link_data']['type'] || null;
 	journeys_utils.isDesktopJourney = data['journey_link_data']['type'] === "desktop";
 	journeys_utils.journeyVariant = data['journey_link_data']['variant'] || null;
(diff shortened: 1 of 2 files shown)


gdeluna-branch:
Closing this since this isn't planned work.

PR #1147, opened 2026-07-28
[patch] ENGMT-1981: preserve BE-supplied body CSS through Journey exit animation

bredmond5:
## Description

When a Journey's iframe CSS comes from the BE (a custom `<style id="branch-iframe-css">` block embedded in the template), any `margin`/`transition` it declares on `body` lives only inside that stylesheet. `animateBannerExit` deletes the whole `#branch-iframe-css` element on close, so the body's margin snaps back to its resting value instead of animating — unlike the SDK-generated CSS path, where the margin is set inline and survives.

This moves just the `margin-top`/`margin-bottom`/`transition` declarations from the BE's parsed `body` rule inline onto `document.body.style` (via a throwaway stylesheet + the browser's own CSS parser, not regex), so they survive the stylesheet removal and combine the BE's margin value with whatever margin the body already had via `calc()` instead of clobbering it. Everything else the BE declares on `body` (background, etc.) is left in the stylesheet only — same as before, it applies while the Journey is up and disappears cleanly on close, rather than leaking permanently onto the page.

Fixes # (no ticket link provided — internal engmt-1981)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] Unit test — `addIframeOuterCSS with a BE-supplied cssIframeContainer` suite in `test/journeys_utils.js` covering: transition copied inline, non-margin/transition declarations (background) explicitly *not* copied inline, margin-top/margin-bottom combined via `calc()` with pre-existing body margin, no-op when the BE CSS has no `body` rule, and cleanup of the temporary parsing stylesheet.
- Full suite: `npx mocha` — 173 passing, 0 failing.

## JS Budget Check

| Files            | Before      | After       |
| -----------      | ----------- | ----------- |
| dist/build.js    | 165850 B    | 166801 B (+951 B) |
| dist/build.min.js| 77166 B     | 77794 B (+628 B)  |

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

## Mentions:

cc @BranchMetrics/saas-sdk-devs for visibility.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/journeys_utils.js
+++ b/src/journeys_utils.js
@@ -340,6 +340,55 @@ journeys_utils.addHtmlToIframe = function(iframe, html, userAgent) {
 	}
 }
 
+function getBodyRuleFromIframeCss(cssIframeContainer) {
+	const tempStyle = document.createElement('style');
+	tempStyle.textContent = cssIframeContainer;
+	document.head.appendChild(tempStyle);
+
+	const cssRules = (tempStyle.sheet && tempStyle.sheet.cssRules) || [];
+	document.head.removeChild(tempStyle);
+
+	for (let i = 0; i < cssRules.length; i++) {
+		if (cssRules[i].selectorText === 'body') {
+			return cssRules[i];
+		}
+	}
+
+	return null;
+}
+
+/***
+ * @function applyBodyCssFromIframeContainer
+ * @param {string} cssIframeContainer - CSS text supplied by the BE for the iframe container
+ * @param {number} bodyMarginTopNumber - margin-top (in px) the body already had before this Journey
+ * @param {number} bodyMarginBottomNumber - margin-bottom (in px) the body already had before this Journey
+ *
+ * Parse the BE's CSS via a throwaway stylesheet
+ * then copies the parsed `body` rule inline so it survives removal of the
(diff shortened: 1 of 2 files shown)


JakeSalgado-Branch:
I was imagining a different approach where the web SDK adds a new class to the page's body, e.g. `branch-banner-is-closing` (and maybe `branch-banner-is-opening` could be nice to have in the future). Then we can add the opening and closing styles we need in the iframe_css set by audience rule.

Adding classes should give us flexibility in the future to change animation styles without needing to change the web SDK again.

Your approach let's us add the preexisting body margin back in, which won't be super straight forward if all the CSS is coming from audience rule.

